PR #2, opened 2021-03-26
[WIP] Remove duplicate nodes but not duplicate values

glyn:


glyn:
Duplicates are not removed according to the spec/

PR #12, opened 2023-01-24
Added cases for filter functions

### LINKED ISSUE #8: Add filter selector tests

gregsdennis:
I think this should round out the current syntax for expressions

Resolves #8 
Relates to spec repo issues [#360](https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/360), [#361](https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/361), [#362](https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/362), [#367](https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/367), and [#368](https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/368).  Most of these have been created as a result of this PR.

### DIFF
--- a/cts.json
+++ b/cts.json
@@ -2139,6 +2139,16 @@
       "result": []
     },
     {
+      "name": "filter, length function, non-array/string arg",
+      "selector" : "$[?length(1)>=2]",
+      "invalid_selector": true
+    },
+    {
+      "name": "filter, length function, result must be compared",
+      "selector" : "$[?length(@.a)]",
+      "invalid_selector": true
+    },
+    {
       "name": "filter, count function",
       "selector" : "$[?count(@..*)>2]",
       "document" : [
@@ -2150,6 +2160,16 @@
         {"a": [1,2,3]},
         {"a": [1], "d":"f"}
       ]
+    },
+    {
+      "name": "filter, count function, non-array/string arg",
+      "selector" : "$[?count(1)>2]",
+      "invalid_selector": true
+    },
+    {
+      "name": "filter, count function, result must be compared",
+      "selector" : "$[?count(@..*)]",
+      "invalid_selector": true
     },
     {
       "name": "filter, match function, found match",
@@ -2177,6 +2197,21 @@
       "selector" : "$[?!match(@.a, '2022-01-[0-9]{2}')]",
       "document" : [{"a": "not a match"}],
       "result": [{"a": "not a match"}]
+    },
+    {
+      "name": "filter, match function, invalid first arg",
+      "selector" : "$[?match(1, '2022-01-[0-9]{2}')]",
+      "invalid_selector": true
+    },
+    {
+      "name": "filter, match function, invalid second arg",
+      "selector" : "$[?match(@.a, 1)]",
+      "invalid_selector": true
+    },
+    {
+      "name": "filter, match function, result cannot be compared",
+      "selector" : "$[?match(@.a, '2022-01-[0-9]{2}')==true]",
+      "invalid_selector": true
     },
     {
       "name": "filter, search function, at the end",
@@ -2219,6 +2254,21 @@
       "selector" : "$[?!search(@.a, '2022-01-[0-9]{2}')]",
       "document" : [{"a": "not a match"}],
       "result": [{"a": "not a match"}]
+    },
+    {
+      "name": "filter, search function, invalid first arg",
+      "selector" : "$[?search(1, '2022-01-[0-9]{2}')]",
+      "invalid_selector": true
+    },
+    {
+      "name": "filter, search function, invalid second arg",
+      "selector" : "$[?search(@.a, 1)]",
+      "invalid_selector": true
+    },
+    {
+      "name": "filter, search function, result cannot be compared",
+      "selector" : "$[?search(@.a, '2022-01-[0-9]{2}')==true]",
+      "invalid_selector": true
     }
   ]
 }
--- a/cts.json
+++ b/cts.json
@@ -2115,6 +2115,85 @@
       "result": [
         {"d": "f"}
       ]
+    },
+    {
+      "name": "filter, length function, string data",
+      "selector" : "$[?length(@.a)>=2]",
+      "document" : [{"a": "ab"}, {"a": "d"}],
+      "result": [
+        {"a": "ab"}
+      ]
+    },
+    {
+      "name": "filter, length function, array data",
+      "selector" : "$[?length(@.a)>=2]",
+      "document" : [{"a": [1,2,3]}, {"a": [1]}],
+      "result": [
+        {"a": [1,2,3]}
+      ]
+    },
+    {
+      "name": "filter, length function, missing data",
+      "selector" : "$[?length(@.a)>=2]",
+      "document" : [{"d": "f"}],
+      "result": []
+    },
+    {
+      "name": "filter, count function",
+      "selector" : "$[?count(@..*)>2]",
+      "document" : [
+        {"a": [1, 2, 3]},
+        {"a": [1], "d": "f"},
+        {"a": 1, "d": "f"}
+      ],
+      "result": [
+        {"a": [1,2,3]},
+        {"a": [1], "d":"f"}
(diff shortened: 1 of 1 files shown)


cts.json
```diff
@@ -2115,6 +2115,160 @@
...
+      "name": "filter, count function",
+      "selector" : "$[?count(@..*)>2]",
+      "document" : [
+        {"a": [1, 2, 3]},
+        {"a": [1], "d": "f"},
+        {"a": 1, "d": "f"}
+      ],
+      "result": [
+        {"a": [1,2,3]},
+        {"a": [1], "d":"f"}
```
glyn, line 2161: Why is the above item in the result? The number of descendants of `@.a` for this item is not greater than 2.

cts.json
```diff
@@ -2115,6 +2115,160 @@
...
+      "invalid_selector": true
+    },
+    {
+      "name": "filter, count function, result must be compared",
+      "selector" : "$[?count(@..*)]",
+      "invalid_selector": true
+    },
+    {
+      "name": "filter, match function, found match",
+      "selector" : "$[?match(@.a, '2022-01-[0-9]{2}')]",
```
glyn, line 2176: Although (possibly except for the single quotes, see below) this is a correct regexp, it's a bit peculiar because it allows dates such as `2022-01-99`. This would be a small distraction to anyone reading the tests. This is exacerbated by a number of tests using the same regexp. Why not avoid this by choosing a different regexp?

cts.json
```diff
@@ -2115,6 +2115,160 @@
...
+      "document" : [{"a": "2022-01-18"}],
+      "result": [
+        {"a":"2022-01-18"}
+      ]
+    },
+    {
+      "name": "filter, match function, don't select match",
+      "selector" : "$[?!match(@.a, '2022-01-[0-9]{2}')]",
```
glyn, line 2184: See https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/367 as to whether regexps may be enclosed in single quotes.

glyn (changes requested):
I like the general content, but a few of the test need changing or need to wait until spec issues are resolved.

gregsdennis:
> a few of the tests need changing or need to wait until spec issues are resolved.

I expected that this would wait for those.

cts.json
```diff
@@ -2115,6 +2115,160 @@
...
+    {
+      "name": "filter, length function, missing data",
+      "selector" : "$[?length(@.a)>=2]",
+      "document" : [{"d": "f"}],
+      "result": []
+    },
+    {
+      "name": "filter, length function, non-array/string arg",
+      "selector" : "$[?length(1)>=2]",
+      "invalid_selector": true
```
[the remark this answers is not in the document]
gregsdennis: That's why `invalid_selector` is set to `true`.

cts.json
gregsdennis: The local path is `@..*`.  That means for the second array item, the following are returned:

- `[1]`
- `1`
- `"f"`

(Both the immediate contents of `@.a` and its descendants are included.

cts.json
gregsdennis: It was familiar-enough.  Happy to change to something else.  Open to proposals.

cts.json
gregsdennis: I looked through the ABNF, and it looks like regex is just a string literal.

I also asked about it.  https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/362 is also releated.

cts.json
```diff
@@ -2115,6 +2115,160 @@
...
+      "name": "filter, length function, array data",
+      "selector" : "$[?length(@.a)>=2]",
+      "document" : [{"a": [1,2,3]}, {"a": [1]}],
+      "result": [
+        {"a": [1,2,3]}
+      ]
+    },
+    {
+      "name": "filter, length function, missing data",
+      "selector" : "$[?length(@.a)>=2]",
```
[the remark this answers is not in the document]
gregsdennis: The referenced issue has been resolved.

cts.json
gregsdennis: Depending on the outcome of #367, we should probably add double-quoted strings as a test case.

cts.json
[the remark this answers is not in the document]
glyn: > Are you saying because `1` is a `Value` this is valid, and it should just return `Nothing`?

Yes.

> 
> If there was something like `$[?length(@.a)>=2]` where the data at `@.a` contained a `1`, then it should return `Nothing`, but an explicitly invalid argument seems like it shouldn't parse.

That would be a spec change which could be considered.

cts.json
glyn: Ah, my mistake. I misread the query.

cts.json
glyn: I would go for something trivial like `a.*` matching `ab` or some such.

gregsdennis: I'll try to rework these, then.

cts.json
glyn: Ah yes. Got it. Thanks.

gregsdennis: Added some double-quoted tests.

cts.json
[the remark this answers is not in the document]
gregsdennis: Please see #368 for `length(1)`.

glyn (changes requested):
Now just waiting for spec issue #368 to be resolved.

gregsdennis:
☝️ as noted by this reference, my implementation now passes all of these tests as well.  I've marked that as dependent upon this PR.

gregsdennis:
@glyn I think this one is good now.

cts.json
```diff
@@ -2115,6 +2120,226 @@
...
+    {
+      "name": "filter, match function, invalid second arg",
+      "selector" : "$[?match(@.a, 1)]",
+      "document" : [{"a": "ab"}],
+      "result": []
+    },
+    {
+      "name": "filter, match function, result cannot be compared",
+      "selector" : "$[?match(@.a, 'a.*')==true]",
+      "invalid_selector": true
```
glyn, line 2255: The spec currently allows functions which return `OptionalNodeOrValue` or one of its subtypes to be used in comparisons. `OptionalBoolean` is a subtype of `OptionalNodeOrValue`. Therefore the result of `match` can be compared.

gregsdennis: `OptionalNodeOrValue` is only defined for functions. Where does the ABNF allow this?

gregsdennis: Okay...

`function-expr` is the only representation of functions in the ABNF (aside from describing the syntax for functions), and that is permissible in all expressions.

However, that means that there can be no type checking during parsing, so what's the point of having a type system at all?

glyn: The ABNF doesn't capture the whole story. There is normative text that describes syntax which is valid in the ABNF but invalid in the spec.

gregsdennis: Yeah, I don't buy this argument.  I've lost so many decisions because what I wanted wasn't expressed in the ABNF.

gregsdennis: I also argue that `OptionalNodeOrValue` shouldn't include `Boolean`.

glyn: Any path with a comparison involving a function that returns an `OptionalNodes` has a type error.

gregsdennis: See https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/387

gregsdennis: Discussion has carried on to https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/404 and https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/405

glyn (approved):
The latest commit LGTM.

gregsdennis:
Thanks. I'll merge this and we can edit later for more spec decisions.

PR #15, opened 2023-01-30
Allow underscores in member name shorthand

f3ath:
Please correct me if I'm wrong, but the standard seems to allow the name shorthands to start from (and consist of) an underscore.

https://ietf-wg-jsonpath.github.io/draft-ietf-jsonpath-base/draft-ietf-jsonpath-base.html#name-syntax-8

```ABNF
child-segment             = bracketed-selection /
                            ("."
                             (wildcard-selector /
                              member-name-shorthand))

bracketed-selection       = "[" S selector *(S "," S selector) S "]"

member-name-shorthand     = name-first *name-char
name-first                = ALPHA /
                            "_"   /            ; _
                            %x80-10FFFF
                              ; any non-ASCII Unicode character
name-char                 = DIGIT / name-first

DIGIT                     =  %x30-39              ; 0-9
ALPHA                     =  %x41-5A / %x61-7A    ; A-Z / a-z
```

### DIFF
--- a/cts.json
+++ b/cts.json
@@ -47,6 +47,17 @@
       ]
     },
     {
+      "name": "name shorthand, underscore",
+      "selector": "$._",
+      "document": {
+        "_": "A",
+        "_foo": "B"
+      },
+      "result": [
+        "A"
+      ]
+    },
+    {
       "name": "name shorthand, symbol",
       "selector": "$.&",
       "invalid_selector": true
@@ -54,16 +65,6 @@
     {
       "name": "name shorthand, number",
       "selector": "$.1",
-      "invalid_selector": true
-    },
-    {
-      "name": "name shorthand, underscore",
-      "selector": "$._",
-      "invalid_selector": true
-    },
-    {
-      "name": "name shorthand, underscore start",
-      "selector": "$._a",
       "invalid_selector": true
     },
     {


f3ath:
@gregsdennis please take a look.

gregsdennis:
I'm not sure this was intended. I thought that we had agreed shorthand names must start with an alpha char.

@glyn can you confirm that intent, please? We may need to update the spec.

glyn:
> I'm not sure this was intended. I thought that we had agreed shorthand names must start with an alpha char.
> 
> @glyn can you confirm that intent, please? We may need to update the spec.

I was intended because we allow a massive range of Unicode characters at the start of a short name, so why not `_` too?

gregsdennis:
Do we have that decision documented anywhere?

f3ath:
@gregsdennis I am not sure what your suggestion is. The expectation part of the tests was that the parser fails. If the underscore is allowed, the expectation does not make sense anymore, hence deleting the tests.

glyn:
> @gregsdennis I am not sure what your suggestion is. The expectation part of the tests was that the parser fails. If the underscore is allowed, the expectation does not make sense anymore, hence deleting the tests.

I think it's good to test edge cases, so those tests should be converted to ones in which parsing does not fail.

f3ath:
Good. I will make the changes.

glyn:
> Good. I will make the changes.

You may wish to drag your feet as there is a JSONPath WG meeting next week and this topic may come up (not guaranteed though!).

cabo:
Please don't drag feet -- I'd like CTS to complete on the ABNF in the current draft.

(Maybe there is confusion with function-name-first, which indeed was decided to be rather restrictive with LCALPHA only.)

gregsdennis:
> @gregsdennis I am not sure what your suggestion is. The expectation part of the tests was that the parser fails. If the underscore is allowed, the expectation does not make sense anymore, hence deleting the tests.

Ah, yes. I didn't see that it's an expected failure.  We need to turn these into successes and verify the return.

gregsdennis:
> which indeed was decided to be rather restrictive with LCALPHA only

@cabo do you know where this decision was recorded?  I can't find it (either way).

Also, seems like lowercase might be a bit too restrictive.  I think ALPHA would be sufficient.

cabo:
> > which indeed was decided to be rather restrictive with LCALPHA only
> 
> @cabo do you know where this decision was recorded? I can't find it (either way).

It was like that since my original 2022-10-30 proposal.
I seem to remember we discussed this briefly in one of the interims around that time, but the minutes are not helping me find that (some of the minutes actually never made it into the proceedings).

> Also, seems like lowercase might be a bit too restrictive. I think ALPHA would be sufficient.

What do we gain by allowing upper case letters in function names?

glyn:
Agreed.

gregsdennis:
> What do we gain by allowing upper case letters in function names?

This isn't about function names; it's about member names, e.g. `$.Abc`.  This seems like it should be perfectly valid.

cabo:
> > What do we gain by allowing upper case letters in function names?
> 
> This isn't about function names; it's about member names, e.g. `$.Abc`. This seems like it should be perfectly valid.

Sure, these are:

``` ABNF
member-name-shorthand = name-first *name-char
name-first          = ALPHA /
                      "_"   /
                      %x80-10FFFF   ; any non-ASCII Unicode character
name-char           = DIGIT / name-first
``` 

I just mentioned the LCALPHA we use for function-name first characters.

gregsdennis:
I think https://github.com/jsonpath-standard/jsonpath-compliance-test-suite/pull/15#issuecomment-1410795175 got us off topic.

The issue is asking specifically about allowing shorthand name selectors starting with underscores.

glyn:
> I think [#15 (comment)](https://github.com/jsonpath-standard/jsonpath-compliance-test-suite/pull/15#issuecomment-1410795175) got us off topic.
> 
> The issue is asking specifically about allowing shorthand name selectors starting with underscores. It was my impression that we had disallowed this.

I'm not sure where that impression came from as the spec is clear on this point and has been since [base-00](https://datatracker.ietf.org/doc/html/draft-ietf-jsonpath-base-00#page-13) in March 2021

gregsdennis:
My recollection is that we had discussed it later.  I guess I was wrong.

@f3ath rather than deleting these tests, they'll need to be made into valid-syntax tests, selecting appropriate data.

cabo:
> I'm not sure where that impression came from as the spec is clear on this point and has been since [base-00](https://datatracker.ietf.org/doc/html/draft-ietf-jsonpath-base-00#page-13) in March 2021

Actually, between -00 and -01 we removed leading digits and hyphens from dot-child-name (-00) and turned it into the current name-first/name-char construction.  But we did not remove "_", so that has indeed been constant.

https://author-tools.ietf.org/iddiff?url1=draft-ietf-jsonpath-base-00&url2=draft-ietf-jsonpath-base-01&difftype=--html

cabo:
> @f3ath rather than deleting these tests, they'll need to be made into valid-syntax tests, selecting appropriate data.

I just changed them into:

``` 
    {
      "name": "name shorthand, underscore",
      "selector": "$._",
      "invalid_selector": false
    },
    {
      "name": "name shorthand, underscore start",
      "selector": "$._a",
      "invalid_selector": false
    },

```

But of course having a document and a result would be better.

gregsdennis:
I don't believe we have any tests with `"invalid_selector": false`.  Either the query is valid and is evaluated or it's invalid.

f3ath:
Updated, please take another look.

cts.json
```diff
@@ -47,23 +47,35 @@
...
-      "invalid_selector": true
+      "name": "name shorthand starting with underscore",
+      "selector": "$._foo",
+      "document": {
+        "_": "A",
+        "_foo": "B"
+      },
+      "result": [
+        "B"
+      ]
```
glyn, line 69: Do we really need two tests for names beginning with underscore?

glyn (approved):
LGTM, but we could safely lose one of the new tests I think.

glyn (approved):

gregsdennis:
@glyn what was the superfluous test? Looks like the cookies were force-pushed wiping the history?

I think it could be valuable having a `$._` and a `$._a`.  I think there's a distinction between a name _starting with_ an underscore and a name _being just_ an underscore.

glyn:
There's certainly a distinction, but IMO the extra $._a test would be
unlikely to fail if the $._ test passed and I would say leading underscores
are going to be vanishingly rare in the wild. But if you think it really
needs to go back in, feel free to make that happen.

f3ath:
Are we good to merge this one? Any change requests?

PR #17, opened 2023-03-08
Add more tests for search() and match().

f3ath:
To cover the cases where the regexp comes from the same document.

I hope you don't mind the autoformat.

### DIFF
--- a/cts.json
+++ b/cts.json
@@ -1680,7 +1680,7 @@
       "result": [
         0,
         1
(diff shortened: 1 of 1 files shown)


gregsdennis (approved):
These pass on my implementation.  I think this is good.  Second pair of eyes from @glyn requested.

glyn (approved):
LGTM, thanks.

Issue #9, opened 2022-07-05
Add tests of non-deterministic behaviour


glyn:
Tests are needed to cover the lack of deterministic ordering of objects. The preferred approach is for such tests to include a set of possible results, any one of which is valid.

glyn:
(The following is a summary of a discussion in PR #60.)

@f3ath [wrote](https://github.com/jsonpath-standard/jsonpath-compliance-test-suite/pull/60#issuecomment-1972077854) (with terminology adjusted to align with RFC 9535):

> Have we considered adding a flag (e.g. `"non-deterministic": true`) to allow arbitrary order of the result array instead of trying to list it in every possible order?

As @gregsdennis [pointed out in PR 60](https://github.com/jsonpath-standard/jsonpath-compliance-test-suite/pull/60#issuecomment-1972116361), the RFC does not allow _arbitrary_ non-determinism in the ordering of the nodelist resulting from a query. (The Working Group discussed increasing the non-determinism in that way, but decided against it.)

[Table 6](https://www.rfc-editor.org/rfc/rfc9535#table-6), [table 12](https://www.rfc-editor.org/rfc/rfc9535#table-12), and [table 16](https://www.rfc-editor.org/rfc/rfc9535#table-16) in the spec include examples where there is _some_ non-determinism without there being arbitrary non-determinism. The limited non-determinism is typically produced by a combination of objects (which are unordered) and arrays (which are ordered) in the query argument. 

> I might not understand the idea fully, but if we want to allow non-deterministic order somewhere deep inside of the complex result object, we could maybe introduce a list of json pointers pointing to the particular elements of the result which are allowed to be non-deterministic.

I considered that possibility, but it's hard to specify simply and each implementation that wants to run the CTS would need to implement it correctly. The approach I've chosen is for non-deterministic tests to list all the allowable result arrays, so implementations simply need to check that the result array is in the list. The burden then falls on the writers of the CTS to correctly list all the allowable result arrays, which isn't _too_ difficult for small examples.

(Another potential approach is to generalise an implementation of the RFC so that it produces a list of all the allowable nodelists and then to use that to prime the CTS. This could be done in Haskell, for example, by using the ["list monad"](https://www.schoolofhaskell.com/school/starting-with-haskell/basics-of-haskell/13-the-list-monad#the-list-monad), but it wouldn't be trivial.)

PR #18, opened 2023-03-16
all values are fine for value args, even if they don't make sense

gregsdennis:
Updates a few test scenarios per https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/431

### DIFF
--- a/cts.json
+++ b/cts.json
@@ -2147,7 +2147,8 @@
     {
       "name": "filter, length function, non-array/string arg",
       "selector" : "$[?length(1)>=2]",
-      "invalid_selector": true
+      "document" : [{"d": "f"}],
+      "result": []
     },
     {
       "name": "filter, length function, result must be compared",
@@ -2239,14 +2240,16 @@
       "result": [{"a": "bc"}]
     },
     {
-      "name": "filter, match function, invalid first arg",
+      "name": "filter, match function, non-string first arg",
       "selector" : "$[?match(1, 'a.*')]",
-      "invalid_selector": true
+      "document" : [{"a": "bc"}],
+      "result": []
     },
     {
-      "name": "filter, match function, invalid second arg",
+      "name": "filter, match function, non-string second arg",
       "selector" : "$[?match(@.a, 1)]",
-      "invalid_selector": true
+      "document" : [{"a": "bc"}],
+      "result": []
     },
     {
       "name": "filter, match function, result cannot be compared",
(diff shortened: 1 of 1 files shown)


cts.json
```diff
@@ -2239,14 +2240,16 @@
...
-      "invalid_selector": true
+      "document" : [{"a": "bc"}],
+      "result": []
     },
     {
-      "name": "filter, match function, invalid second arg",
+      "name": "filter, match function, non-string second arg",
       "selector" : "$[?match(@.a, 1)]",
-      "invalid_selector": true
+      "document" : [{"a": "bc"}],
```
f3ath, line 2251: Change to `[{"a": 1}]` maybe?

cts.json
```diff
@@ -2320,14 +2323,16 @@
...
+      "result": []
     },
     {
-      "name": "filter, search function, invalid second arg",
+      "name": "filter, search function, non-string second arg",
       "selector" : "$[?search(@.a, 1)]",
-      "invalid_selector": true
+      "document" : [{"a": "bc"}],
```
f3ath, line 2334: Same suggestion as above

cts.json
gregsdennis: Why?  The point is that it doesn't select the item because `1` is not a string.

gregsdennis: But the expression isn't governed by the underlying engine.  It's our own syntax, and we don't have type coercion.

Also for the type system as it currently is, `1` and `"1"` are two unrelated values within the same type: `ValueType`.

gregsdennis: I think it would be better to have a test for `1=="1"` to test for that specifically.

gregsdennis: The second case only applies when there is coercion, and the tests I added in #23 ensure that doesn't happen.  It's not necessary to check that here.

glyn: Some general comments:

* Each test should aim to test one thing.

* We may be able to borrow some tests from the JSONPath comparison project, or at least get inspiration from there, particularly for tests that check that non-standard behaviour does not occur. (@cburgmer may care to comment.)

glyn (approved):
LGTM, thanks.

## Meta comment

Please note that I don't have sufficient bandwidth to give the CTS much attention. I'm delighted that there are now at least two others involved. (I'm still concerned about the lack of a good approach to https://github.com/jsonpath-standard/jsonpath-compliance-test-suite/issues/9, which I think needs implementing for the CTS to be sufficient. Maybe we can take up discussion of that in the issue.)

PR #19, opened 2023-03-16
add tests for value func

gregsdennis:


f3ath (approved):

gregsdennis:
@f3ath I've updated these slightly.  Looks like I missed the function typedness verifications.

PR #22, opened 2023-03-19
Disallow non-singular queries in comparison expressions

f3ath:


cts.json
```diff
@@ -2122,6 +2122,30 @@
         {"d": "f"}
       ]
     },
+    {
+      "name": "filter, non-singular query in comparison, slice",
+      "selector" : "$[?@[0:0]==0]",
+      "document" : [[0, 1], [0, 1, 2]],
```
gregsdennis, line 2128: If the selector is invalid, we don't need a document.  These can be removed.

gregsdennis (changes requested):
These look good.  Just a bit too much info.

gregsdennis (approved):

PR #25, opened 2023-03-23
Add a test for nested filters

f3ath:


cts.json
```diff
@@ -2174,6 +2174,15 @@
       "selector" : "$[?@.a[*].a==0]",
       "invalid_selector": true
     },
+    {
+      "name": "filter, nested",
+      "selector": "$.a[?@>count($.b[?@>5])]",
```
gregsdennis, line 2179: Is there another way to make a nested query that doesn't include a function?  We only want to test one thing here.  This tests

- nested queries
- (non-singular) queries as function args
- `count()` function

gregsdennis (changes requested):

cts.json
gregsdennis: How about `$[?@[?@.a=='foo']]`?  In an array of arrays, this would return any array that contained an object that contains an `a` property with a value of "foo".

Data:

```json
[
  [1, {"a": "foo"}],
  [2, {"b": "foo"}],
  [3. {"a": "bar"}]
]
```

Result:
```json
[
  [1, {"a": "foo"}]
]
```

f3ath: Good point, will update.

gregsdennis (approved):
Yep, that looks good.

PR #27, opened 2023-03-24
Add JSON Schema validation, fixes #26

### LINKED ISSUE #26: CTS tests?
It seems that we might benefit from some automated quality gates for the CTS itself. One thing I can imagine is validation against JSON Schema, that would catch typos at least.

f3ath:
@gregsdennis please take a look.  Tested it on [my fork](https://github.com/f3ath/jsonpath-compliance-test-suite/pull/1). Not sure what to put in the `$id`.

P.S. I just learned that you're one of the JSON Schema authors :)

Edit:
Resolves #26 (you have to put the "resolves/fixes/etc" text in the body of the opening comment for the automatic close to work)

### DIFF
--- a/cts.schema.json
+++ b/cts.schema.json
@@ -1,0 +1,80 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://raw.githubusercontent.com/jsonpath-standard/jsonpath-compliance-test-suite/main/cts.schema.json",
+  "title": "Compliance Test Suite",
+  "description": "JSONPath Compliance Test Suite",
+  "type": "object",
+  "properties": {
+    "tests": {
+      "description": "Individual test cases",
+      "type": "array",
+      "items": {
+        "$ref": "#/$defs/test_case"
+      }
+    }
+  },
+  "required": [
+    "tests"
+  ],
+  "$defs": {
+    "test_case": {
+      "type": "object",
+      "properties": {
+        "name": {
+          "$ref": "#/$defs/test_case_name"
+        },
+        "selector": {
+          "$ref": "#/$defs/selector"
+        },
+        "document": {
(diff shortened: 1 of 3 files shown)


cts.schema.json
```diff
@@ -0,0 +1,80 @@
...
+      "type": "array",
+      "items": {
+        "$ref": "#/$defs/test_case"
+      }
+    }
+  },
+  "required": [
+    "tests"
+  ],
+  "$defs": {
```
gregsdennis, line 19: I think it's fine for all of these to just be inlined.  We're not really gaining anything from having these references.  If anything it hurts readability because we're having to read a def name then scan for what it means.

`$defs` is really good for deduplication, but since we're not defining the same schema for multiple places in the data, it doesn't matter.

gregsdennis (approved):
This will work as is.  Nice job.  Just the one comment, but I think that's more a matter of preference.

The `$id` is fine.  It's long, but it's fine.

gregsdennis:
Can you submit a change to the CTS that would break the schema, please?  That should trigger a build failure.

gregsdennis:
Looks like it might need to be.  I don't have permission to do that on this repo.  @glyn?

image

f3ath:
@gregsdennis you can see it in action here https://github.com/f3ath/jsonpath-compliance-test-suite/pull/1

gregsdennis:
@glyn do you have any issues with this?  I think it's a good addition that will help keep the CTS content valid.

glyn:
> @glyn do you have any issues with this? I think it's a good addition that will help keep the CTS content valid.

@gregsdennis No, looks good to me. I've made you an admin of this repo so you can do anything that's needed.

gregsdennis:
> I've made you an admin of this repo so you can do anything that's needed.

Yeah, just wanted your thoughts before I merged.

PR #31, opened 2023-04-09
Test suite split PoC

### LINKED ISSUE #21: Split into several files?
As the test suite grows it might be worth splitting into several files to improve maintainability. As least, maybe, extract the standard functions tests into individual ones. I could craft a PR for that.

f3ath:
An attempt to split the CTS into manageable pieces. The individual test suites are in the `tests` directory. The `build.sh` reads the `tests` recursively and places the result into `cts.json`. The directory/file names are prepended to the test names and the underscores are replaced with spaces.

The resulting `cts.json` is included with the PR. It is possible to build the `cts.json` automatically in a GitHub action later.

Since the resulting `cts.json` is different from the original (the names of some tests have changed, the tests are reordered), here's a simple script to compare the before/after:
```javascript
const fs = require('fs');
const tests = JSON
    .parse(fs.readFileSync(0).toString())
    .tests;
tests.forEach(test => delete test.name);
tests.sort((a, b) => JSON.stringify(a).localeCompare(JSON.stringify(b)));
console.log(JSON.stringify(tests, null, 2));
```
It can be run as `cat cts.json | node strip_names.js | sha256sum`.

Resolves #21

### DIFF
--- a/.github/workflows/validate_json_schema.yaml
+++ b/.github/workflows/validate_json_schema.yaml
@@ -10,5 +10,5 @@ jobs:
       - name: Validate CTS against JSON Schema
         uses: cardinalby/schema-validator-action@v2
         with:
-          file: 'cts.json'
+          file: 'tests/**/*.json|cts.json'
           schema: 'cts.schema.json'
--- a/build.js
+++ b/build.js
@@ -87,7 +87,7 @@ function readTestsFromDir(dir, relativePath = []) {
         const secondIsDir = isDir(second);
         if (firstIsDir && !secondIsDir) return 1;
         if (!firstIsDir && secondIsDir) return -1;
-        return first.localeCompare(second);
+        return first.localeCompare(second, 'en', {sensitivity: 'base'});
     }
 
     /**
--- a/build.js
+++ b/build.js
@@ -1,0 +1,126 @@
+const fs = require('fs');
+const path = require('path');
+/**
+ * The file extension of test files.
+ * @type {string}
+ */
+const fileExtension = '.json';
+/**
+ * The number of spaces to use for indentation in the output.
+ * @type {number}
+ */
+const jsonIndent = 2;
+/**
+ * The separator to use when combining test names.
+ * @type {string}
+ */
+const testNameSeparator = ', ';
+
+build();
+
+/**
+ * Build a combined test suite from a directory of test files.
+ * The output is written to stdout.
+ */
+function build() {
+    if (process.argv.length < 3) {
+        console.error('Usage: node build.js <test-folder-path>');
+        process.exit(1);
+    }
+
+    const testsFolder = process.argv[2];
+    const tests = readTestsFromDir(testsFolder);
+    const cts = {'tests': tests};
+    console.log(JSON.stringify(cts, null, jsonIndent));
+    console.error(`Wrote ${tests.length} tests to stdout.`);
+}
+
+
+/**
+ * Read all test files from a directory and its subdirectories.
+ * The directory name is prepended to the test name.
+ * @param dir {string} - The directory to read.
+ * @param relativePath {string[]} - The path to the directory.
+ * @returns {*[]} - An array of test objects.
+ */
+function readTestsFromDir(dir, relativePath = []) {
+    const files = fs.readdirSync(dir);
+    files.sort(filesBeforeDirs);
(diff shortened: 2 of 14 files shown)


build.js
```diff
@@ -0,0 +1,126 @@
...
+                .replace('_', ' ');
+            const tests = readTestsFromFile(fullPath).map(prependToTestName(testNamePrefix));
+            return allTests.concat(tests);
+        }
+
+        return allTests;
+    }
+
+    /**
+     * Sort entries so that directories are listed after files.
```
glyn, line 80: Would it be true to say that this function also sorts the files and the directories into some sort of lexicographic order? (Not sure what localeCompare does. In particular, I'm not sure if this would sort differently depending on the locale - probably not desirable. It would be good to have a sort that is stable across all environments.)

build.sh
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env sh
+set -e
+path=$(dirname "$0")
+node build.js "$path/tests" > "$path/cts.json"
```
glyn, line 4: I don't know much about node. How heavy a dependency is it, in terms of size and frequency of required upgrade?

.github/workflows/validate_json_schema.yaml
```diff
@@ -10,5 +10,5 @@ jobs:
       - name: Validate CTS against JSON Schema
         uses: cardinalby/schema-validator-action@v2
         with:
-          file: 'cts.json'
+          file: 'tests/**/*.json|cts.json'
```
glyn, line 13: The fact that the individual JSON files are test suites in their own right, although a clever feature, somewhat cuts against the requirement that all implementers should consume the same set of tests.

Having a less sophisticated way of combining the individual JSON files into cts.json could avoid having to install node. It might be possible to use just a shell script or, failing that, a shell script that depends on just jq.

OTOH I can see that having the individual files conform to the same schema is a bonus for early detection of syntax errors.

glyn (commented):
This is a very neat solution and a great PoC. Please see my detailed comments.

.github/workflows/validate_json_schema.yaml
f3ath: The implementers should consume the cts.json and nothing else.

JavaScript seemed the most natural language to operate with JSON. Operating with pieces of JSON as strings would be error prone IMO. So we will have to use some tool for that. Using an actual programming language (instead of a query language) seems beneficial should we decide to add more logic (e.g. version generation) to the build script in the future. 

Early detection is the main reason why I left the structure the same.

build.js
f3ath: Yes, that's how I understand it. A hard-coded locale can be provided to this method so it won't depend on the environment. I will make a update soon.

build.sh
f3ath: I can't see nodejs being a substantial obstacle for anyone wishing to mess with the CTS. It's available for pretty much every popular OS and the distribution size is about 30 MB. https://nodejs.org/en/download

f3ath: My idea was that the build script would run on the server, so you wouldn't need to install node locally. You would just make changes to the files in `tests/` directory and submit a PR. GitHub actions would then take care of validating the changes and updating the `cts.json`.

glyn (approved):
Looks like a great starting point.

build.sh
glyn: I'm not sure that's the fastest way to get feedback if someone has an implementation and is working on the CTS. I suspect they'd install node and run build.sh to test their changes locally.

build.js
glyn: Thanks. It's kind of strange to bring locale into the code at all, isn't it? Maybe node forces that.

.github/workflows/validate_json_schema.yaml
glyn: > The implementers should consume the cts.json and nothing else.

Let's not ignore JSONPath implementors who also need to update the CTS.

gregsdennis: > Let's not ignore JSONPath implementors who also need to update the CTS.

That's what a readme or a CONTRIBUTING file is for.  I doubt implementors will have a hard time with it.

f3ath:
Any more comments on this?

glyn:
> Any more comments on this?

It's been quiet, so let's merge.

PR #35, opened 2023-04-26
Add an edge case test that `.` matches \\u2029 and \\u2028

f3ath:
I-Regexp follows the [XSD-2](https://www.w3.org/TR/2004/REC-xmlschema-2-20041028/#regexs) which states that the equivalent character class for `.` is `[^\r\n]`. Some programming languages (e.g. Javascript, Dart) treat `.` differently, in particular it won't match Unicode chars `\u2029` and `\u2028`. This PR introduces a corresponding edge case test.

### DIFF
--- a/cts.json
+++ b/cts.json
@@ -3455,6 +3455,36 @@
       ]
     },
     {
+      "name": "functions, match, dot matcher on \\u2028",
+      "selector": "$[?match(@, '.')]",
+      "document": [
+        " ",
+        "\r",
+        "\n",
+        true,
+        [],
+        {}
+      ],
+      "result": [
+        " "
+      ]
+    },
+    {
+      "name": "functions, match, dot matcher on \\u2029",
+      "selector": "$[?match(@, '.')]",
+      "document": [
+        " ",
+        "\r",
+        "\n",
+        true,
+        [],
+        {}
+      ],
+      "result": [
+        " "
+      ]
+    },
+    {
       "name": "functions, match, result cannot be compared",
       "selector": "$[?match(@.a, 'a.*')==true]",
       "invalid_selector": true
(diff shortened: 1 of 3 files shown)


cts.json
```diff
@@ -3454,6 +3454,36 @@
         "a𐄁b"
       ]
     },
+    {
+      "name": "functions, match, dot matcher on \\u2028",
```
glyn, line 3458: Would it be possible to use \u2028 in the JSON document?

f3ath: Not sure I understand the question. Do you mean removing the second `\`? That would make the test name indistinguishable from the other one.

glyn: No, I was talking about using `\u2028` in the `"document"` member.

f3ath: It is how it's defined in the document, you can see it in the source file. I'm not sure if it would be possible to keep it as `\uXXXX` in the compiled `cts.json`.

gregsdennis:
My implementation (which uses the .Net regex engine) is returning the `\r` and `\n` as well.

```
Name:     functions, match, dot matcher on \u2028
Selector: $[?match(@, '.')]
Document: ["\u2028","\r","\n",true,[],{}]
Result:   ["\u2028"]
Results:   null
IsValid:  True

Actual (values): ["\u2028","\r","\n"]

Actual (serialized):
{
  "Matches": [
    {
      "Value": "\u2028",
      "Location": "$[0]"
    },
    {
      "Value": "\r",
      "Location": "$[1]"
    },
    {
      "Value": "\n",
      "Location": "$[2]"
    }
  ],
  "Error": null
}
```

gregsdennis (approved):
After some googling I figured out how to add the lookahead exclusions, and the tests pass for me now.

hiltontj:
Was able to fix this to get things passing in `serde_json_path` again with https://github.com/hiltontj/serde_json_path/pull/92. Thanks for surfacing this one @f3ath!

Issue #40, opened 2023-07-15
ABNF says whitespace should be allowed between root and selector


graza:
The spec defines the ABNF for a sequence of segments in a query as allowing whitespace before each segment.  There is a number of tests in tests/whitespace/selectors.json that mark queries with such syntax as invalid selectors.  The ABNF is as follows:

```
jsonpath-query      = root-identifier segments
segments            = *(S segment)
root-identifier     = "$"
segment             = child-segment / descendant-segment
```

gregsdennis:
So it does. Los like I missed this one. Can you create a PR fixing these test cases?

f3ath:
Spaces in between segments look weird to my taste. Wondering how widespread this style is "in the wild".

gregsdennis:
The argument for it was being able to place segments on separate lines.  I like to think of things like StackOverflow where you're trying to explain what each part does.

graza:
> Can you create a PR fixing these test cases?

Will do.

PR #41, opened 2023-07-17
Fix for #40 Whitespace allowed between root and selector

graza:
Modified tests/whitespace/selectors.json to change those tests that were marking paths with whitespace between the root '$' and the selector as invalid.  Rebuilt and committed cts.json as part of this. #40

### DIFF
--- a/cts.json
+++ b/cts.json
@@ -5568,62 +5568,130 @@
     {
       "name": "whitespace, selectors, space between root and bracket",
       "selector": "$ ['a']",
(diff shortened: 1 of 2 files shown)


gregsdennis:
I've made changes to my implementation to allow whitespace in these locations.  Let me run against your updated suite... a few minutes...

gregsdennis (approved):
These pass with the changes I made to allow this whitespace.  Thanks for the PR!

PR #42, opened 2023-07-20
Cover whitespaces in selectors within function args

f3ath:


glyn (approved):
LGTM thanks.

f3ath:
@gregsdennis any objections to this?

gregsdennis (approved):
Yep looks like this is covered in the ABNF as well.  Thanks.

PR #43, opened 2023-07-22
Negative tests for non-singular queries

f3ath:
Related discussion: https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/481


gregsdennis (approved):
These all pass in my implementation.  Thanks.

f3ath:
Fixed the "less-or-equal" cases. This is ready for merge now.

PR #46, opened 2023-08-28
Cover whitespace in and around filter selectors

### LINKED ISSUE #44: Whitespace between "?" and logical-expr
It looks like whitespace is allowed between the question mark of a filter selector and its expression.

```plain
filter-selector     = "?" S logical-expr
```

I don't think there are any test cases covering this yet.

jg-rp:
Cover whitespace in and around filter selectors. Closes #44.

I note that there is arguably nothing special about whitespace around parenthesized expressions, and choosing to test cases such as "space between question mark and parenthesized expression" might be considered a little arbitrary, compared to any other filter expression not surrounded by parentheses.

Similarly, one might consider there to be unnecessary overlap between "space between bracket and question mark" etc. and cases from `whitespace/selectors.json`.  I would understand if these test cases need to be removed.

### DIFF
--- a/cts.json
+++ b/cts.json
@@ -3968,6 +3968,326 @@
       "invalid_selector": true
     },
     {
+      "name": "whitespace, filter, space between question mark and expression",
+      "selector": "$[? @.a]",
+      "document": [
+        {
+          "a": "b",
+          "d": "e"
+        },
+        {
+          "b": "c",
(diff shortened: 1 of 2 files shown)


gregsdennis (approved):
These all pass for me!  Thanks.

f3ath (approved):

PR #47, opened 2023-08-28
Cover multiple selectors inc filters and wildcards

### LINKED ISSUE #45: Multiple selectors including wildcard-selector and filter-selector
If I've understood the spec correctly (that's a big "if"), bracketed-selection can include wildcard and filter selectors, potentially producing duplicate nodes.

```plain
selector            = name-selector  /
                      wildcard-selector /
                      slice-selector /
                      index-selector /
                      filter-selector

bracketed-selection = "[" S selector *(S "," S selector) S "]"
```

For example, these would be valid queries.

```plain
$.store.book[*, 0, 1, 1:3]
$.store.book[*, *, *]
$.store.book[?@.isbn, 1, 2, 3, *]
```

I don't believe this is covered by the test suite yet.

jg-rp:
Cover multiple selectors including filters and wildcards. Closes #45.

### DIFF
--- a/cts.json
+++ b/cts.json
@@ -239,6 +239,115 @@
       ]
     },
     {
+      "name": "basic, multiple selectors, duplicate index",
+      "selector": "$[1,1]",
+      "document": [
+        0,
+        1,
+        2,
+        3,
+        4,
+        5,
+        6,
+        7,
+        8,
+        9
+      ],
+      "result": [
+        1,
+        1
+      ]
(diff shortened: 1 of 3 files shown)


gregsdennis (approved):
These all work for me, and it looks like everything is basically covered.  (Not combinatorically, but sufficiently.)

f3ath (approved):

PR #49, opened 2023-09-11
Cover empty name selectors

jg-rp:
Cover name selectors that are an empty string literal. Like `$['']` and `$[""]`.


gregsdennis (approved):
These pass for my implementation.

PR #50, opened 2023-09-11
Cover equality of numbers inc literal fractions and exponents

jg-rp:
For reference, this is the relevant part of the grammar.

```plain
literal             = number / string-literal /
                      true / false / null
number              = (int / "-0") [ frac ] [ exp ] ; decimal number
frac                = "." 1*DIGIT                  ; decimal fraction
exp                 = "e" [ "-" / "+" ] 1*DIGIT    ; decimal exponent
```

### DIFF
--- a/cts.json
+++ b/cts.json
@@ -2177,6 +2177,227 @@
       ]
     },
     {
+      "name": "filter, equals number, zero and negative zero",
+      "selector": "$[?@.a==-0]",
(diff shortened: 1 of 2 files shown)


gregsdennis (approved):
The ones with `+` don't pass for me, but I think that's a bug.  The tests look right.  (fixed now)

PR #51, opened 2023-09-12
Cover expressions as function arguments

jg-rp:
These additional test cases don't make a lot of real world sense, but short of defining one or more non-standard function extensions specifically for testing, it's the best I could come up with.

Relevant grammar:

```plain
function-expr       = function-name "(" S [function-argument
                         *(S "," S function-argument)] S ")"
function-argument   = literal /
                      filter-query / ; (includes singular-query)
                      logical-expr /
                      function-expr

logical-expr        = logical-or-expr
logical-or-expr     = logical-and-expr *(S "||" S logical-and-expr)
                        ; disjunction
                        ; binds less tightly than conjunction
logical-and-expr    = basic-expr *(S "&&" S basic-expr)
                        ; conjunction
                        ; binds more tightly than disjunction

basic-expr          = paren-expr /
                      comparison-expr /
                      test-expr

paren-expr          = [logical-not-op S] "(" S logical-expr S ")"
                                        ; parenthesized expression
```

### DIFF
--- a/cts.json
+++ b/cts.json
@@ -3820,6 +3820,48 @@
       "invalid_selector": true
     },
     {
+      "name": "functions, arguments, parenthesized expression",
+      "selector": "$[?match((@.a), ('a.*'))]",
+      "document": [
+        {
+          "a": "ab"
+        },
+        {
+          "a": "ba"
+        }
+      ],
+      "result": [
+        {
+          "a": "ab"
+        }
+      ]
+    },
+    {
+      "name": "functions, arguments, function expression",
+      "selector": "$.values[?match(@.a, value($..['regex']))]",
+      "document": {
+        "regex": "a.*",
+        "values": [
+          {
+            "a": "ab"
+          },
+          {
+            "a": "ba"
+          }
+        ]
+      },
+      "result": [
+        {
+          "a": "ab"
+        }
+      ]
+    },
+    {
+      "name": "functions, arguments, type mismatch, value type and logical type",
+      "selector": "$[?length(@.a==@.b)>2]",
+      "invalid_selector": true
+    },
+    {
       "name": "functions, count, count function",
       "selector": "$[?count(@..*)>2]",
       "document": [
--- a/tests/functions/arguments.json
+++ b/tests/functions/arguments.json
@@ -1,0 +1,25 @@
+{
+  "tests": [
+    {
+      "name": "parenthesized expression",
+      "selector": "$[?match((@.a), ('a.*'))]",
+      "document" : [{"a": "ab"}, {"a": "ba"}],
+      "result": [
+        {"a": "ab"}
+      ]
+    },
+    {
+      "name": "function expression",
+      "selector": "$.values[?match(@.a, value($..['regex']))]",
+      "document" : {"regex": "a.*", "values": [{"a": "ab"}, {"a": "ba"}]},
+      "result": [
+        {"a": "ab"}
+      ]
+    },
+    {
+      "name": "type mismatch, value type and logical type",
+      "selector": "$[?length(@.a==@.b)>2]",
+      "invalid_selector": true
+    }
+  ]
+}


cts.json
```diff
@@ -3819,6 +3819,58 @@
...
+      },
+      "result": [
+        {
+          "a": "ab"
+        }
+      ]
+    },
+    {
+      "name": "functions, arguments, comparison expression",
+      "selector": "$[?length(@.a==@.b)>2]",
```
gregsdennis, line 3861: I'm not sure what this one is trying to do.  `@.a==@.b` yields a `LogicalType`, but `length()` takes a `ValueType`.  This path is semantically invalid (syntactically valid, though).  It shoudl error.

gregsdennis (changes requested):

cts.json
jg-rp: The intention was to test that a comparison expression is a valid function argument, within the confines of the standard function extensions. 

I see that if `LogicalTrue` and `LogicalFalse` are not equal to JSON `true` and `false`, then the result of `@.a==@.b` would not be a `ValueType`, making the query invalid.

If `LogicalType` can be used in place of a JSON Boolean (implicitly converting it to a `ValueType`?), I would expect `length()` to return `Nothing`, and the comparison between `Nothing` and `2` to result in `LogicalFalse`.

I'm not sure, and I can always define some mock functions that do accepts a `LogicalType` parameter, outside of the CTS.

gregsdennis: We opted to separate the result of a boolean expression from the JSON literals `true` and `false` and recognize that the JSON literals are just tokens that don't carry any meaning.  There's good reason to do this, and I'll let you peruse the issues for more, but the main was that if we didn't separate these concepts, then test expressions would be ambiguous.

For example, consider the complete expression `@.foo`.  Is this testing whether `foo` exists, or whether `foo` contains a JSON `true`?  Unless we are explicit that the JSON `true` is separate from logical true, this can't be determined.

I expect that this test case should generate an error.  @glyn can you confirm, please?  What's the rule for syntactically correct but semantically incorrect paths?

jg-rp: Thank you, @gregsdennis , that makes sense. I've provisionally marked that test case as invalid and given it a more relevant name.

[Section 2.4.9](https://datatracker.ietf.org/doc/html/draft-ietf-jsonpath-base-20#section-2.4.9) has some examples that cover of this kind of thing. I had missed that before.

cts.json
```diff
@@ -3819,6 +3819,48 @@
...
+      "result": [
+        {
+          "a": "ab"
+        }
+      ]
+    },
+    {
+      "name": "functions, arguments, function expression",
+      "selector": "$.values[?match(@.a, value($..['regex'])]",
+      "document": {
```
f3ath, line 3842: For an invalid selector, the document is irrelevant and should be omitted.

jg-rp: Unless I've misunderstood (quite possible :smiley:), `$.values[?match(@.a, value($..['regex'])]` looks like a valid query, and the given result is correct when applied to that document.

1. `$..['regex']` evaluates to a nodelist, of `NodesType`, which happens to contain a single node.
2. `value()` accepts a `NodesType` argument and returns a `ValueType`. In this case `value($..['regex'])` evaluates to `"a.*"`, because its argument contains a single node.
3. `match()` is called with the `ValueType` implicitly converted from the singular query `@.a`, and the `ValueType`  (`"a.*"`) from the call to `value()`.

gregsdennis: The idea is good, but the path is invalid.  It needs another `)`

```jsonpath
$.values[?match(@.a, value($..['regex']))]
```

Tested at https://json-everything.net/json-path.

jg-rp: Thanks. Looks like I have some issues with unbalanced bracket detection.

cts.json
```diff
@@ -3819,6 +3819,48 @@
       "selector": "$[-1:-10:-231584178474632390847141970017375815706539969331281128078915168015826259279872]",
       "invalid_selector": true
     },
+    {
+      "name": "functions, arguments, parenthesized expression",
+      "selector": "$[?match((@.a), ('a.*'))]",
```
f3ath, line 3824: I'm not sure this is correct. Here's how I see it. The parens around the arguments can only be used in a `logical-exp`, which building blocks (apart from the parens, `&&`, `||`) are `comparison-expr` and `test-expr`. So `@.a` and `a.*` are `test-expr` meaning they are `LogicalType`. But `match()` only accepts `ValueType` arguments, therefore the expression is not well-typed and the implementation must raise an error.

gregsdennis: I see where you're going with this. I'm mobile right now, but it can be tested at https://json-everything.net/json-path.

f3ath: @gregsdennis 
image

f3ath: One can argue that it could be better not to allow `logical-expr` in
```abnf
function-argument   = literal /
                      filter-query / ; (includes singular-query)
                      logical-expr /
                      function-expr
 ```

Comment (f3ath) [reply cts.json:3824]: FWIW my implementation did not allow parens in function args until now. This is a useful test, there's just no such standard function that can be used in it. In my implementation I introduced a non-standard logical function to have this case covered.

Comment (gregsdennis) [reply cts.json:3824]: > One can argue that it could be better not to allow logical-expr in...

We don't want to preclude the possibility of someone registering such a function if it proves useful.

Comment (cabo) [reply cts.json:3824]: > `$[?match((@.a), ('a.*'))]`

```
$ pbpaste | jpt -
/Users/cabo/lib/ruby/gems/3.2.0/gems/jpt-0.1.2/lib/jpt.rb:84:in `from_jp': Expected one of " ", "\t", "\n", "\r", "==", "!=", "<=", ">=", "<", ">" at line 1, column 23 (byte 23) after [?match((@.a), ('a.*' (RuntimeError)
Expected one of " ", "\t", "\n", "\r", "==", "!=", "<=", ">=", "<", ">" at line 1, column 23 (byte 23):
$[?match((@.a), ('a.*'))]
~~~~~~~~~~~~~~~~~~~~~~^
```
Clearly, ('a.*') is not a logical-expr, so the ABNF grammar forbids this.

> `$[?match((@.a), 'a.*')]`

```
$ pbpaste | jpt -
["$", ["filt", ["func", "match", ["@", "a"], "a.*"]]]
```

Comment (gregsdennis) [reply cts.json:3824]: My implementation still doesn't like `$[?match((@.a), 'a.*')]`.  Removing the parentheses around `@.a` works, though.

I think @f3ath's analysis is correct that `(@.a)` is interpreted as a general expression (yielding `LogicalType`).  Specifically, it doesn't meet the requirements of a singular query.

The relevant spec section is [2.4.3](https://www.ietf.org/archive/id/draft-ietf-jsonpath-base-21.html#section-2.4.3), toward the bottom:

> (2) The arguments of the function expression are well-typed when each argument of the function can be used for the declared type of the corresponding parameter, according to one of the following conditions:
> 
> - ...
> - When the declared type of the parameter is ValueType and the argument is one of the following:
>     - A value expressed as a literal.
>     - A singular query. In this case:
>       - If the query results in a nodelist consisting of a single node, the argument is the value of the node.
>       - If the query results in an empty nodelist, the argument is the special result Nothing.

Now, it could be argued that `(@.a)` can be determined to be a singular query _at runtime_, after the query runs and results are returned, but simple syntactic analysis doesn't allow `(@.a)` where a `ValueType` is expected.

However, making this argument means that the query's well-typedness can't be determined until runtime, which (to me) seems to violate section [2.1](https://www.ietf.org/archive/id/draft-ietf-jsonpath-base-21.html#section-2.1):

> A JSONPath implementation MUST raise an error for any query which is not well-formed and valid. The well-formedness and the validity of JSONPath queries are independent of the JSON value the query is applied to. **No further errors relating to the well-formedness and the validity of a JSONPath query can be raised during application of the query to a value.** This clearly separates well-formedness/validity errors in the query from mismatches that may actually stem from flaws in the data.

Comment (f3ath) [reply cts.json:3824]: @gregsdennis Can you please clarify why the well-typedness can't be determined at compile time? Given a function whose arguments' types are known at compile type, that each argument is exactly one of `ValueType`, `NodesType`, or ` LogicalType`, and considering the type (Nodes to Value and Nodes to Logical) it should be possible IMO.

Comment (gregsdennis) [reply cts.json:3824]: I was just saying that if you were to wait until runtime to determine if a query were singular, then you wouldn't be able to determine well-typedness until runtime, either, since well-typedness can depend on whether a query is singular.

---

The primary problem is that `(@.a)` is _syntactically_ a filter expression.  Expressions return `LogicalType`.  (In the context of an expression, a path on its own is an existence test, which returns `LogicalType`.)

Furthermore, we don't (and probably shouldn't) have a conversion _from_ `LogicalType`.

This means that the expression isn't well-formed as `match()` isn't receiving correctly-typed arguments.

Comment (f3ath) [reply cts.json:3824]: Ah, understood.

Comment (gregsdennis) [reply cts.json:3824]: @cabo does this mean that the test you [showed](https://github.com/jsonpath-standard/jsonpath-compliance-test-suite/pull/51#discussion_r1335207281) has a bug, or am interpreting this wrong?

Comment (cabo) [reply cts.json:3824]: Well, the problem is that a reasonable parser implementation will discard the () around `@.a`, so the fact that this was syntactically a logicaltype is not visible to the semantic processing.

If I force this by doing one of

``` 
$[?match(!(@.a), 'a.*')]
$[?match(!@.a, 'a.*')]
$[?match((@.a && @.a), 'a.*')]
```

I get one of

```
*** Cannot use ["not", ["@", "a"]] with declared_type logical for required type value in ["func", "match", ["not", ["@", "a"]], "a.*"]
*** Cannot use ["and", ["@", "a"], ["@", "a"]] with declared_type logical for required type value in ["func", "match", ["and", ["@", "a"], ["@", "a"]], "a.*"]
```

Giving pure parentheses a type-coercing semantics is a flaw that overly complicated languages such as C++ have, but that I didn't know we had put into JSONPath.  Now I have to think how to integrate this into jpt's parser/ast generator.

jg-rp:
I'm inclined to scrap this PR in favour of a new one, "Cover well-typedness of standard functions", which can add to existing test file in `tests/functions/`.

That would leave handling of redundant parentheses and covering well-typedness of functions accepting `LogicalType` arguments as a separate issues.

Any objections?

gregsdennis:
I don't have problems with adding commits to this.

If you create a new PR, please link back to this one to capture the conversation.

cts.json
gregsdennis: > reasonable parser implementation will discard the () around `@.a`

I'm not convinced of this.  Reasonability is subjective.

cabo: s/reasonable parser/parser written by someone who has done expression language parsers before/

The AST gets a bit ugly when you need to include the parens as a node.
Of course, it can be done.

gregsdennis:
@jg-rp do you have the time/ability to continue with this?

jg-rp:
I've gone ahead an created PR #53 covering the use of function expressions as function arguments for some of the standard function extensions.

I'll open new issues shortly related to the extra parentheses discussion here, and to summarise the theoretically valid combinations of nested function expressions that **can't** be covered using standard function extensions alone.

PR #53, opened 2023-12-20
Cover function expressions as function parameters

jg-rp:
These additional test cases cover using function expressions as function parameters, plus an additional `length()` case [mentioned in the spec](https://datatracker.ietf.org/doc/html/draft-ietf-jsonpath-base-21#name-examples-7).

Note that it's not currently possible to test `count()` and `value()` with function expressions as parameters using standard function extensions alone. None of the standard function extensions return a _NodesType_.

This PR partially replaces #51. See also the conversation on PR #51.

### DIFF
--- a/cts.json
+++ b/cts.json
@@ -2398,6 +2398,32 @@
       ]
     },
     {
+      "name": "filter, equals, special nothing",
+      "selector": "$.values[?length(@.a) == value($..c)]",
+      "document": {
+        "c": "cd",
+        "values": [
+          {
+            "a": "ab"
+          },
+          {
+            "c": "d"
+          },
+          {
+            "a": null
+          }
+        ]
+      },
+      "result": [
+        {
+          "c": "d"
+        },
+        {
+          "a": null
+        }
+      ]
+    },
+    {
       "name": "index selector, first element",
       "selector": "$[0]",
       "document": [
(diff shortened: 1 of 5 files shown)


f3ath (approved):
I like this addition. @gregsdennis care to merge?

gregsdennis (changes requested):
These look good, but I'd like to see some examples where `value()` returns `Nothing` (the nodelist is either empty or contains multiple elements).

For example, with this data

```json
{
  "c": "cd",
  "values": [
    {
      "a": "ab"
    },
    {
      "c": "d"
    },
    {
      "a": null
    }
  ]
}
```

`value(@.a)` returns `Nothing` for `{"c": "d"}` and `null` for `{"a": null"}`.  `length()` should then return `Nothing` for both of these values.

`value($..c)` returns `Nothing` because there are multiple matching locations.

Finally, `Nothing == Nothing` returns true, meaning that the last two elements are returned in the final result:

```json
[
  {
    "c": "d"
  },
  {
    "a": null
  }
]
```

gregsdennis:
@glyn can you verify my logic ☝️ please?   This is what my lib currently does.

glyn:
> @glyn can you verify my logic ☝️ please? This is what my lib currently does.

LGTM

jg-rp:
> These look good, but I'd like to see some examples where `value()` returns `Nothing` (the nodelist is either empty or contains multiple elements).

Hi @gregsdennis, I've added a couple more cases which I believe cover the behaviour demonstrated by your example.

cts.json
```diff
@@ -2397,6 +2397,32 @@
...
+            "a": null
+          }
+        ]
+      },
+      "result": [
+        {
+          "c": "d"
+        },
+        {
+          "a": "null"
```
gregsdennis, line 2422: This `"null"` needs to be JSON `null` (unquoted)

jg-rp: Fixed :+1:

gregsdennis (approved):
These work for me.

PR #56, opened 2024-02-25
Add tests for incorrect filter usage

Marcono1234:
This mainly covers using filter on primitives, which should have no result, see [RFC 9535 section 2.3.5.2](https://www.rfc-editor.org/rfc/rfc9535#name-semantics-7):
> The filter selector works with arrays and objects exclusively. [...] Applied to a primitive value, it selects nothing

(This was inspired by https://github.com/hiltontj/serde_json_path/issues/49; CC @hiltontj)

Please let me know if any of the tests are incorrect, or if I should adjust them or split this into separate pull requests. Any feedback is appreciated!

### DIFF
--- a/cts.json
+++ b/cts.json
@@ -510,6 +510,26 @@
       ]
     },
     {
+      "name": "filter, existence without selector",
+      "selector": "$[?@]",
+      "document": {
+        "a": {},
+        "b": 1
+      },
+      "result": [
+        {},
+        1
+      ]
+    },
+    {
+      "name": "filter, existence test on primitive (not supported)",
+      "selector": "$.a[?@]",
+      "document": {
+        "a": 1
+      },
+      "result": []
+    },
+    {
       "name": "filter, equals string, single quotes",
       "selector": "$[?@.a=='b']",
       "document": [
@@ -1843,6 +1863,43 @@
           42
         ]
       ]
+    },
+    {
+      "name": "filter, on primitive (not supported)",
+      "selector": "$.a[?@ == 1]",
+      "document": {
+        "a": 1
+      },
+      "result": []
+    },
+    {
+      "name": "filter, on primitive object member value (@ refers to member values, not the object)",
+      "selector": "$[?@.a == 1]",
+      "document": {
+        "a": 1
+      },
+      "result": []
+    },
+    {
+      "name": "filter, name selector on array",
+      "selector": "$[?@[\"0\"] == 5]",
+      "document": [
+        [
+          5,
+          6
+        ]
+      ],
+      "result": []
+    },
+    {
+      "name": "filter, index selector on object",
+      "selector": "$[?@[0] == 5]",
+      "document": [
+        {
+          "0": 5
+        }
+      ],
+      "result": []
     },
     {
       "name": "filter, relative non-singular query, index, equal",
@@ -4094,6 +4151,22 @@
           "d": "f"
         }
       ],
+      "result": []
+    },
+    {
+      "name": "functions, length, filter on primitive array element (@ refers to array elements, not the array)",
+      "selector": "$[?length(@)==1]",
+      "document": [
+        1
+      ],
+      "result": []
+    },
+    {
+      "name": "functions, length, filter on primitive object member value (@ refers to member values, not the object)",
+      "selector": "$[?length(@)==1]",
+      "document": {
+        "a": 1
+      },
       "result": []
     },
     {
--- a/tests/filter.json
+++ b/tests/filter.json
@@ -17,6 +17,18 @@
       ]
     },
     {
+      "name": "existence without selector",
+      "selector": "$[?@]",
+      "document": {"a": {}, "b": 1},
+      "result": [{}, 1]
+    },
+    {
+      "name": "existence test on primitive (not supported)",
+      "selector": "$.a[?@]",
+      "document": {"a": 1},
+      "result": []
+    },
+    {
       "name": "equals string, single quotes",
       "selector" : "$[?@.a=='b']",
       "document" : [{"a": "b", "d": "e"}, {"a":"c", "d": "f"}],
@@ -477,6 +489,30 @@
         [0, 1, 2],
         [42]
       ]
+    },
+    {
+      "name": "on primitive (not supported)",
+      "selector": "$.a[?@ == 1]",
+      "document": {"a": 1},
+      "result": []
+    },
+    {
+      "name": "on primitive object member value (@ refers to member values, not the object)",
+      "selector": "$[?@.a == 1]",
+      "document": {"a": 1},
+      "result": []
+    },
+    {
+      "name": "name selector on array",
+      "selector": "$[?@[\"0\"] == 5]",
+      "document": [[5, 6]],
+      "result": []
+    },
+    {
+      "name": "index selector on object",
+      "selector": "$[?@[0] == 5]",
+      "document": [{"0": 5}],
+      "result": []
     },
     {
       "name": "relative non-singular query, index, equal",
--- a/tests/functions/length.json
+++ b/tests/functions/length.json
@@ -53,6 +53,18 @@
       "result": []
     },
     {
+      "name": "filter on primitive array element (@ refers to array elements, not the array)",
+      "selector": "$[?length(@)==1]",
+      "document": [1],
+      "result": []
+    },
+    {
+      "name": "filter on primitive object member value (@ refers to member values, not the object)",
+      "selector": "$[?length(@)==1]",
+      "document": {"a": 1},
+      "result": []
+    },
+    {
       "name": "result must be compared",
       "selector" : "$[?length(@.a)]",
       "invalid_selector": true


cts.json
```diff
@@ -509,6 +509,26 @@
...
+        "a": {},
+        "b": 1
+      },
+      "result": [
+        {},
+        1
+      ]
+    },
+    {
+      "name": "filter, existence test on primitive (not supported)",
```
hiltontj, line 525: I think the wording "not supported" is a bit misleading. The wording in the standard is:

> Applied to a primitive value, it selects nothing [...]

Therefore, I suggest the test name:

```json
"filter, existence test on primitive, selects nothing"
```

I would apply this suggestion to other, similarly named tests.

Also, note that I did not include brackets, which seem to violate the general naming convention used throughout the test suite.

cts.json
```diff
@@ -1844,6 +1864,43 @@
         ]
       ]
     },
+    {
+      "name": "filter, on primitive (not supported)",
```
hiltontj, line 1868: Suggestion:

```json
"filter, on primitive (selects nothing)"
```

cts.json
```diff
@@ -1844,6 +1864,43 @@
...
+      "name": "filter, on primitive object member value (@ refers to member values, not the object)",
+      "selector": "$[?@.a == 1]",
+      "document": {
+        "a": 1
+      },
+      "result": []
+    },
+    {
+      "name": "filter, name selector on array",
+      "selector": "$[?@[\"0\"] == 5]",
```
hiltontj, line 1885: Should use single quotes for name selectors to reduce line noise:

```json
"$[?@['0'] == 5]"
```

hiltontj (commented):
Thank you @Marcono1234 for opening the issue and for CC'ing me. I have left a couple of comments for suggestion, but feel that approval should come from one of the other contributors that has worked on this test suite more extensively than I.

One general comment that I wanted to make was this: I had thought of adding a test case back when https://github.com/hiltontj/serde_json_path/issues/49 was resolved, but felt the problem in that case was more due to a logical error in `serde_json_path`'s code, as opposed to some violation of the standard.

So, that at least was the excuse I made for not taking the time to do what you have done here 🙂. Having these test cases will help prevent other future implementors from making the same logical errors I did, so I would give a 👍.

Marcono1234:
Thanks for the feedback! I hope I addressed everything as you had it in mind. As part of that I removed the "@ refers to ..." from the names, hopefully it is still clear why the queries should have no result, respectively don't work as one might expect.

To the maintainers: Please feel free to squash the commits on merge, or let me know if I should squash them for a cleaner Git history.

tests/filter.json
```diff
@@ -478,6 +490,30 @@
...
     },
+    {
+      "name": "on primitive, selects nothing",
+      "selector": "$.a[?@ == 1]",
+      "document": {"a": 1},
+      "result": []
+    },
+    {
+      "name": "on primitive object member value, selects nothing",
+      "selector": "$[?@.a == 1]",
```
gregsdennis, line 501: This and the following three tests seem to be checking that a relative path (`@<...>`) can have the same path syntax as a global path (`$<...>`).  If that's something we're testing, that's fine, but I'd argue that we'd need to almost replicate the entire suite for relative paths.

gregsdennis (changes requested):
There's some duplication of existing tests in here.  I'd like to be sure we have good reason to add them and that the test cases are minimal.

Also, I'd like @glyn's opinion.

FWIW, these all pass in my implementation.

tests/filter.json
```diff
@@ -478,6 +490,30 @@
         [42]
       ]
     },
+    {
+      "name": "on primitive, selects nothing",
+      "selector": "$.a[?@ == 1]",
+      "document": {"a": 1},
+      "result": []
+    },
```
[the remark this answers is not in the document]
Marcono1234: > This could be improved since the `.a` isn't necessary.

It seems that would be the first test then with primitive top-level value. I had created #57 to propose in general adding tests for that.

> Also, I don't think having a filter here improves the test. It could be any selector.

But those are (at least in the specification) two separate cases which independently say that only arrays and objects are supported, so maybe it would be good to have separate tests?

tests/functions/length.json
```diff
@@ -52,6 +52,18 @@
...
       "result": []
     },
+    {
+      "name": "filter on primitive array element, selects nothing",
+      "selector": "$[?length(@)==1]",
+      "document": [1],
+      "result": []
+    },
+    {
+      "name": "filter on primitive object member value, selects nothing",
```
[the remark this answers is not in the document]
Marcono1234: You are right; I mainly added this to make sure no implementation misinterprets `@` as being the whole JSON object (instead of just the member value `1`).

tests/filter.json
Marcono1234: This one was based on https://github.com/hiltontj/serde_json_path/issues/49 where the reporter assumed `@` refers to the JSON object instead of the member value.

And the tests "name selector on array" and "index selector on object" are supposed to cover the same issue in `serde_json_path` where the underlying bug was that an unexpected type (which should select nothing) erroneously always passed the filter. I assume such a bug pattern could affect other libraries as well.

tests/filter.json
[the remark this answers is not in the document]
gregsdennis: > so maybe it would be good to have separate tests

As noted in other comments, we _do_ have those tests.

tests/functions/length.json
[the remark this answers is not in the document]
gregsdennis: I think that can be tested without the `length()` function.

tests/functions/length.json
```diff
@@ -52,6 +52,18 @@
       "document" : [{"d": "f"}],
       "result": []
     },
+    {
+      "name": "filter on primitive array element, selects nothing",
+      "selector": "$[?length(@)==1]",
+      "document": [1],
+      "result": []
+    },
```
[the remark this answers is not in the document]
Marcono1234: Same as with https://github.com/jsonpath-standard/jsonpath-compliance-test-suite/pull/56#discussion_r1501902004; the intention was to make sure no library misinterprets `@` as representing the JSON array (instead of its elements).

tests/filter.json
gregsdennis: > where the reporter assumed @ refers to the JSON object instead of the member value.

Two things here:

1. We shouldn't be interested in testing what an end user expects, even (especially) if that expectation is wrong.  We should be testing _implementation_ behavior per the spec.  If the implementation is doing it right and the user expected something different, then the solution is to correct the user.  We don't need a test for it.
2. If we _are_ going to test this, then we should test for interpreting `@` directly; we don't need the `.a`.  `"$[?@ == 1]" would be a much more targeted test.

tests/functions/length.json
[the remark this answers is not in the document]
Marcono1234: Could you provide an example for such a test please?

My intention was that if a library misinterprets `@` as being the array, then `length(@) == 1` passes and it erroneously selects the array as result. Arguably this is not so much about the `length` function.

Maybe using `@==@` could work here though (and this could be in `filter.json` then)? The expected result is `[1]`, but a library which erroneously considers `@` as referring to the JSON array would have `[[]]` or `[[], 1]` or similar as result?

tests/filter.json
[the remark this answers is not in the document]
gregsdennis: (I thought we had those tests for primitives but we only have tests for e.g. "index selector against object".)

tests/functions/length.json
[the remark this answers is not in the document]
gregsdennis: `$[?@ == 1]` against either an object or an array works just fine to test this.

tests/filter.json
Marcono1234: > 1. We shouldn't be interested in testing what an end user expects, even (especially) if that expectation is wrong.  We should be testing _implementation_ behavior per the spec.

Fair point. But I think here if the document is `{"a": 1}`, then it is necessary to have `.a`. If you used `$[?@ == 1]`, then `[1]` is the correct result.

But if you add `.a` (i.e. `?@.a == 1`), then you should get no result. However, libraries might get this wrong if they silently ignore the `.a` because it cannot be applied to a non-object. And with https://github.com/hiltontj/serde_json_path/issues/49 there is a precedent for that exact issue.

tests/functions/length.json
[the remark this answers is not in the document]
Marcono1234: But that would not detect `@` being misinterpreted as referring also to the array (instead of just its items). It would in all cases only have the `1` as result.

tests/filter.json
gregsdennis: I view this differently.  The _test_ should be that `@` represents the `1`.  This is the simple test that verifies the implementation has the correct behavior.

Separately, we test that `.a` doesn't apply to a non-object (which is your primitive tests).

That `.a` can't be applied to a non-object in the context of `@` is merely a _consequence_ of the correct behavior and doesn't specifically need a test.

tests/functions/length.json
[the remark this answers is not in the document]
gregsdennis: > It would in all cases only have the 1 as result.

Yes, exactly, and in doing so, it correctly tests that the implementation is interpreting `@` not as the container but as the value within it.

tests/filter.json
hiltontj: `singular-query`s, i.e., relative paths with `@`, are defined separately in the standard ABNF. So, this leads implementations to have separate parsing, AST representations, evaluation mechanisms, etc. - as was the case for `serde_json_path`.

Therefore, I think there is value in having tests that ensure the behaviour of queries against primitives for _both_ absolute _and_ relative paths. Relative paths can only appear in filters, so having separate tests for those requires queries like @Marcono1234 has used here.

---

I suppose that the motivation for this test case is that one implementor (me) made a logic error in the handling of singular queries, so, having this test would prevent another from doing the same. It is not possible to come up with tests for every conceivable logic error, or misinterpretation of JSONPath, but I don't see the harm in having a test that guards against one that _did happen_ (and the `.a` was relevant in that case).

gregsdennis: > `singular-querys`, i.e., relative paths with @, are defined separately in the standard ABNF.

While true, `@`-paths can also be existence tests, which can handle all selector constructs.  I believe (on mobile now; can verify letter) that we have some treats that ensure only singular paths are used in comparisons, though.  If not, we should add them in another PR.

> I suppose that the motivation for this test case is that one implementor (me) made a logic error in the handling of singular queries, so, having this test would prevent another from doing the same.

Okay. That this was an implementation error wasn't understood.  I thought this was a user that expected something incorrect.

However, I still maintain, that this test is overly complex and can be broken down into smaller constituent pieces that still achieve the goal of verifying proper operation.

cts.json
```diff
@@ -4096,6 +4153,22 @@
       ],
       "result": []
     },
+    {
+      "name": "functions, length, filter on primitive array element, selects nothing",
+      "selector": "$[?length(@)==1]",
```
glyn, line 4158: This test could be made broader, thus:
```suggestion
      "selector": "$[?length(@)>=0]",
```

cts.json
```diff
@@ -4096,6 +4153,22 @@
...
+      "document": [
+        1
+      ],
+      "result": []
+    },
+    {
+      "name": "functions, length, filter on primitive object member value, selects nothing",
+      "selector": "$[?length(@)==1]",
```
glyn, line 4166:
```suggestion
      "selector": "$[?length(@)>=0]",
```

cts.json
glyn: Also, it would be good to add a test for the comparison of two `Nothing` values if we don't already, e.g.:
```
$[?length(@)==length(@)]
```
should select all the elements of an array argument.

glyn (approved):
LGTM with some suggestions. I can't comment on whether tests are duplicates as I'm not keeping track of all the tests.

cts.json
gregsdennis: https://github.com/jsonpath-standard/jsonpath-compliance-test-suite/blob/main/tests%2Ffilter.json

We have `Nothing` equality.

Marcono1234:
Thanks for all your reviews! Sorry that I did not approach this in a very structured way, and also mixed multiple test cases in this pull request.

@gregsdennis, if you prefer feel free to add tests in a more organized and simplified way, then I will close this pull request. Otherwise, should I only address @glyn's comments, or try to organize and simplify the tests more as you suggested @gregsdennis?

cts.json
gregsdennis: @glyn I think these changes are moot given the conversation I've raised.  What these tests are trying to verify can be accomplished with several more targeted tests.

gregsdennis:
I think @glyn and I (at a minimum) should agree on the approach for this.  I'm still of the mind that these tests are too complex and the desired behavior can be covered with a few more targeted test cases.

glyn:
> I think @glyn and I (at a minimum) should agree on the approach for this. I'm still of the mind that these tests are too complex and the desired behavior can be covered with a few more targeted test cases.

@Marcono1234 May I suggest that you close this PR (or convert it into a draft PR) and then submit some small PRs, each targetting one case? I hope that won't be much more work and it would mean that we can quickly merge the non-contentious PRs. (If you have a better suggestion, I'm all ears!)

tests/filter.json
hiltontj: Sorry for taking so long to get back...

> While true, @-paths can also be existence tests

This escaped me, thanks for pointing that out. I was incorrect there. Relative path `!=` singular query. The singular query defined in the ABNF is only relevant to comparisons. Relative paths in existence tests or as function arguments can be non-singular.

PR #58, opened 2024-02-26
Add GitHub workflow for updating `cts.json`

### LINKED ISSUE #33: CI to check that cts.json has been built
To avoid the situation where cts.json gets out of step with the built version, we could define a CI check that builds cts.json and makes sure it equals the version checked in.

Marcono1234:
Resolves #33

For every push to `main` the new workflow runs `build.sh` and afterwards commits & pushes the changes for `cts.json`, if any.

Example workflow runs:
- [No change](https://github.com/Marcono1234/jsonpath-compliance-test-suite/actions/runs/8052973356)
- [Change](https://github.com/Marcono1234/jsonpath-compliance-test-suite/actions/runs/8052985119) and commit https://github.com/Marcono1234/jsonpath-compliance-test-suite/commit/b01ffbb6ad4f3c9f24d1a4bcf0be2ff168905415
- [Change & unexpected change](https://github.com/Marcono1234/jsonpath-compliance-test-suite/actions/runs/8053015649)

Limitations:
- The push event for updating `cts.json` does not trigger any other workflows, see [action documentation](https://github.com/stefanzweifel/git-auto-commit-action/tree/master?tab=readme-ov-file#commits-made-by-this-action-do-not-trigger-new-workflow-runs)
- This might need additional configuration if `main` is a protected branch, see [action documentation](https://github.com/stefanzweifel/git-auto-commit-action/tree/master?tab=readme-ov-file#push-to-protected-branches)
(might depend on who is the "actor" triggering the workflow; so maybe if a member who is allowed to push to protected branches merged the pull request to `main`, then it works without additional configuration?)

Feedback is appreciated!

### DIFF
--- a/.github/workflows/build_cts_json.yaml
+++ b/.github/workflows/build_cts_json.yaml
@@ -43,8 +43,8 @@ jobs:
     # Important: The push event will not trigger any other workflows, see
     # https://github.com/stefanzweifel/git-auto-commit-action?tab=readme-ov-file#commits-made-by-this-action-do-not-trigger-new-workflow-runs
     - name: Commit & push changes
-      # Only run for push events (on `main` branch); don't run for pull requests
-      if: github.event_name == 'push'
+      # Only run for push events on `main` branch
+      if: github.event_name == 'push' && github.ref == 'refs/heads/main'
       uses: stefanzweifel/git-auto-commit-action@v5
       with:
         commit_message: 'Update `cts.json`'
--- a/.github/workflows/build_cts_json.yaml
+++ b/.github/workflows/build_cts_json.yaml
@@ -1,15 +1,18 @@
-name: Update 'cts.json'
+name: Build 'cts.json'
 on:
   push:
     branches:
       - main
+  pull_request:
 
 jobs:
-  update-cts:
+  build-cts:
     runs-on: ubuntu-latest
 
     permissions:
       # Allow the job to push the changed file to the repository
+      # For pull requests from forks this seems to be implicitly changed to 'read', as desired, see
+      # https://docs.github.com/en/actions/security-guides/automatic-token-authentication#how-the-permissions-are-calculated-for-a-workflow-job
       contents: write
 
     steps:
@@ -19,7 +22,7 @@ jobs:
       id: setup-node
       uses: actions/setup-node@v4
       with:
-        node-version: 20
+        node-version: 'lts/*'
 
     - name: Run build
       run: ./build.sh
@@ -40,6 +43,8 @@ jobs:
     # Important: The push event will not trigger any other workflows, see
     # https://github.com/stefanzweifel/git-auto-commit-action?tab=readme-ov-file#commits-made-by-this-action-do-not-trigger-new-workflow-runs
     - name: Commit & push changes
+      # Only run for push events (on `main` branch); don't run for pull requests
+      if: github.event_name == 'push'
       uses: stefanzweifel/git-auto-commit-action@v5
       with:
         commit_message: 'Update `cts.json`'
--- a/.github/workflows/update-cts-json.yaml
+++ b/.github/workflows/update-cts-json.yaml
@@ -1,0 +1,46 @@
+name: Update 'cts.json'
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  update-cts:
+    runs-on: ubuntu-latest
+
+    permissions:
+      # Allow the job to push the changed file to the repository
+      contents: write
+
+    steps:
+    - uses: actions/checkout@v4
+
+    - name: Setup Node.js
+      id: setup-node
+      uses: actions/setup-node@v4
+      with:
+        node-version: 20
+
+    - name: Run build
+      run: ./build.sh
+
+    # To be safe, verify that either there are no changes or only `cts.json` has changed
+    - name: Verify no unexpected changes
+      run: |
+        # Check for changes to any file other than `cts.json`,
+        # see https://stackoverflow.com/a/29374503
+        # Note that this does not detect new untracked files
+        if ! git diff --exit-code --quiet -- . ':!cts.json'; then
+          echo "Unexpected changes:"
+          git diff -- . ':!cts.json'
+          exit 1
+        fi
+
+    # Commit and push changes; has no effect if the file did not change
+    # Important: The push event will not trigger any other workflows, see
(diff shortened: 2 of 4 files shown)


.github/workflows/update-cts-json.yaml
```diff
@@ -0,0 +1,46 @@
...
+      contents: write
+
+    steps:
+    - uses: actions/checkout@v4
+
+    - name: Setup Node.js
+      id: setup-node
+      uses: actions/setup-node@v4
+      with:
+        node-version: 20
```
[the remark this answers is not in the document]
gregsdennis: I don't think it's possible.  I have some actions that I've had to keep updated manually.

.github/workflows/update-cts-json.yaml
```diff
@@ -0,0 +1,46 @@
+name: Update 'cts.json'
+on:
+  push:
+    branches:
+      - main
```
gregsdennis, line 5: I think it'd be good to run the first two steps at the PR level but just not do the commit.  You can prevent that step from running with an `if:` condition that checks to see if we're on `main`.  ([example](https://github.com/gregsdennis/json-everything/blob/master/.github/workflows/dotnet-core.yml#L83))

.github/workflows/update-cts-json.yaml
```diff
@@ -0,0 +1,46 @@
...
+        fi
+
+    # Commit and push changes; has no effect if the file did not change
+    # Important: The push event will not trigger any other workflows, see
+    # https://github.com/stefanzweifel/git-auto-commit-action?tab=readme-ov-file#commits-made-by-this-action-do-not-trigger-new-workflow-runs
+    - name: Commit & push changes
+      uses: stefanzweifel/git-auto-commit-action@v5
+      with:
+        commit_message: 'Update `cts.json`'
+        file_pattern: 'cts.json'
```
gregsdennis, line 46: `if:` would go in here.

gregsdennis (changes requested):

.github/workflows/update-cts-json.yaml
Marcono1234: By first two steps you mean "Run build" and "Verify no unexpected changes", right?

Should that maybe be a separate job or workflow then (would complicate the setup though), because for pull requests it should probably not have `contents: write` permission. However, the [documentation](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#how-the-permissions-are-calculated-for-a-workflow-job) says 'write' permissions are changed to 'read' for pull requests from forks.

.github/workflows/update-cts-json.yaml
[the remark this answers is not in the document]
Marcono1234: That might work, see https://github.com/actions/setup-node?tab=readme-ov-file#supported-version-syntax. Would probably then be `lts/*` for the latest LTS version. Should I change it?

.github/workflows/update-cts-json.yaml
gregsdennis: > By first two steps you mean "Run build" and "Verify no unexpected changes", right?

Yes.  These should run as part of the PR.

> Should that maybe be a separate job or workflow then

I find it easier to just include as one workflow.  For example, the workflow I linked to runs building a test badge during a PR to make sure that it works.  But for `master` it also pushes that to a gist that I can then reference in my readme.

Marcono1234: I have updated the workflow now; I hope that is what you had in mind.

.github/workflows/build_cts_json.yaml
```diff
@@ -0,0 +1,51 @@
...
+          git diff -- . ':!cts.json'
+          exit 1
+        fi
+
+    # Commit and push changes; has no effect if the file did not change
+    # Important: The push event will not trigger any other workflows, see
+    # https://github.com/stefanzweifel/git-auto-commit-action?tab=readme-ov-file#commits-made-by-this-action-do-not-trigger-new-workflow-runs
+    - name: Commit & push changes
+      # Only run for push events (on `main` branch); don't run for pull requests
+      if: github.event_name == 'push'
```
gregsdennis, line 47: Wouldn't this trigger if someone pushes to a branch?  I think we need to check branch name instead.

Marcono1234: I think the `branches: - main` condition at the top of this file should limit this only to `main`.

Though I can also remove the condition there (so that it is instead run for all branches) and instead add the branch name check here.

gregsdennis: > I think the `branches: - main` condition at the top of this file should limit this only to `main`.

Those are an "any of" set, so any activity on the PR branch would still trigger this (that's how I have mine set up).

The only way to get this to only run on `main` is to check the branch.

---

We should also add `workflow_dispatch:` so that we can manually run it.  It's just a convenience.

Marcono1234: > Those are an "any of" set, so any activity on the PR branch would still trigger this

But I think in that case the `github.event_name` wouldn't be `push` but `pull_request`? (more specifically `pull_request` with [activity type `synchronize`](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request)?)


> We should also add `workflow_dispatch:` so that we can manually run it. It's just a convenience.

I guess I can add that, but assuming this workflow works as expected then (at least for the `main` branch) there would be no need for it I assume. And the `if` condition here would then also have to allow the `workflow_dispatch`. Should I add it nonetheless?

gregsdennis: > more specifically `pull_request` with activity type `synchronize`?

No, I think `pull-request` only has to do with pull request actions: opening, approving, comments, etc.

Regardless, I know it works the way I have it set up.  Please just do that.

Marcono1234: `synchronize` is [described as](https://docs.github.com/en/webhooks/webhook-events-and-payloads?actionType=synchronize#pull_request):
> A pull request's head branch was updated. For example, the head branch was updated from the base branch or new commits were pushed to the head branch.

In my short testing in my fork it looks that is also what was triggered when I made changes on the pull request branch. I assume that you are right, when the pull request branch is in the same repository (and not in a fork) it would also trigger a `push` event, but probably for the pull request branch then, and not `main`.

Either way, having a second check here does not hurt and makes it less error-prone, so thanks for the suggestion. I have changed it now.

As side note: This pull request has "Allow edits" enabled, so feel free to directly make any changes to it, if that is easier.

gregsdennis (approved):

gregsdennis:
Thanks @Marcono1234

PR #60, opened 2024-02-29
Introduce some tests of non-determinism

glyn:
This is not necessarily an exhaustive set of tests of the non-deterministic features of RFC 9535. However, it establishes a convention for how to write such tests.

A non-deterministic test has a "results" key instead of a "result" key. The value corresponding to "results" is an array or possible results. If an implementation produces any one of these, the test should pass. If an implementation produces any other result, the test should fail.

Ref https://github.com/jsonpath-standard/jsonpath-compliance-test-suite/issues/9

### DIFF
--- a/cts.schema.json
+++ b/cts.schema.json
@@ -66,7 +66,6 @@
         },
         {
           "required": [
-            "name",
             "invalid_selector"
           ],
           "properties": {
--- a/cts.schema.json
+++ b/cts.schema.json
@@ -74,6 +74,8 @@
     },
     "test_case_results": {
       "type": "array",
+      "items": {"$ref": "#/$defs/test_case_result"},
+      "minItems": 2,
       "description": "An array of possible expected results of applying the selector to the document, each element of which contains all the matched values"
     },
     "selector": {
--- a/cts.schema.json
+++ b/cts.schema.json
@@ -50,6 +50,7 @@
             "result"
           ],
           "properties": {
+            "invalid_selector": false,
             "results": false
           }
         },
@@ -59,13 +60,20 @@
             "results"
           ],
           "properties": {
+            "invalid_selector": false,
             "result": false
           }
         },
         {
           "required": [
+            "name",
             "invalid_selector"
-          ]
+          ],
+          "properties": {
+            "document": false,
+            "result": false,
+            "results": false
+          }
         }
       ]
     },
--- a/cts.schema.json
+++ b/cts.schema.json
@@ -48,13 +48,19 @@
           "required": [
             "document",
             "result"
-          ]
+          ],
+          "properties": {
+            "results": false
+          }
         },
         {
           "required": [
             "document",
             "results"
-          ]
+          ],
+          "properties": {
+            "result": false
+          }
         },
         {
           "required": [
--- a/cts.schema.json
+++ b/cts.schema.json
@@ -32,6 +32,9 @@
         "result": {
           "$ref": "#/$defs/test_case_result"
         },
+        "results": {
+          "$ref": "#/$defs/test_case_results"
+        },
         "invalid_selector": {
           "$ref": "#/$defs/invalid_selector"
         }
@@ -62,6 +65,10 @@
     "test_case_result": {
       "type": "array",
       "description": "The expected result of applying the selector to the document, contains all the matched values"
+    },
+    "test_case_results": {
+      "type": "array",
+      "description": "An array of possible expected results of applying the selector to the document, each element of which contains all the matched values"
     },
     "selector": {
       "description": "The JSONPath selector",
--- a/cts.schema.json
+++ b/cts.schema.json
@@ -52,6 +52,12 @@
         },
         {
           "required": [
+            "document",
+            "results"
+          ]
+        },
+        {
+          "required": [
             "invalid_selector"
           ]
         }
--- a/README.md
+++ b/README.md
@@ -20,3 +20,7 @@ Many editors support that file natively. Others (such as VS code) require a plug
 To add or modify a test suite, edit the corresponding file in the `tests` directory.
 To generate `cts.json`, run the `build.sh` located in the root folder. Do not modify `cts.json` directly.
 More details are available in the [Contributor Guide](./CONTRIBUTING.md).
+
+### Non-determinism
+
+Where the spec allows non-deterministic results for a given testcase, the testcase should specify an array of all the valid results (each of which is itself an array representing the resultant nodelist from the query) in the "results" member (and should not specify a "result" member).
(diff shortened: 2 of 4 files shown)


cts.schema.json
```diff
@@ -63,6 +72,10 @@
       "type": "array",
       "description": "The expected result of applying the selector to the document, contains all the matched values"
     },
+    "test_case_results": {
+      "type": "array",
```
glyn, line 76: @gregsdennis Is it possible to specify that the type of "test_case_results" is an array of "test_case_result"?

glyn:
@cabo and @hiltontj: since you pass the CTS, you may be interested in this PR.

cts.schema.json
gregsdennis: Yes.  Add `"items": {"$ref": "#/$defs/test_case_result"}`.

gregsdennis:
I'll need to update my runner to account for this change. I can work on that tomorrow.

cts.schema.json
glyn: @gregsdennis Excellent! Done. Also, I made the size of the array at least two. (If there's only one possible result, the testcase should be using "result" instead.)

cts.schema.json
```diff
@@ -32,6 +32,9 @@
         "result": {
           "$ref": "#/$defs/test_case_result"
         },
+        "results": {
+          "$ref": "#/$defs/test_case_results"
+        },
```
glyn, line 37: @gregsdennis I would like to specify that "result" and "results" are mutually exclusive, but I don't think JSON Schema supports that. Right?

gregsdennis: You can.  Where `result` is required, add `"properties": { "results": false } }`, and vice versa.  This makes all valid in `results` invalid.  It's effectively a "don't use this".

glyn: @gregsdennis Perfect. Thanks.

cts.schema.json
```diff
@@ -45,12 +48,31 @@
           "required": [
             "document",
             "result"
-          ]
+          ],
+          "properties": {
+            "invalid_selector": false,
```
glyn, line 53: @gregsdennis I took the liberty of tightening this up, even though it's not needed for the current PR.

cts.schema.json
```diff
@@ -45,12 +48,31 @@
...
+            "result": false
+          }
         },
         {
           "required": [
             "invalid_selector"
-          ]
+          ],
+          "properties": {
+            "document": false,
```
glyn, line 72: @gregsdennis I took the liberty of tightening this up, even though it's not needed for the current PR.

gregsdennis (approved):
Updating my runner wasn't too difficult.  All tests pass.

It's going to be "fun" filling out the rest of the tests...

f3ath:
Have we considered adding a flag (e.g. `"indeterministic": true`) to allow arbitrary order of the result array instead of trying to list it in every possible order?

gregsdennis:
The non-determinism is actually only for object selection.  As you can see in the new test on line 480, while the values from the object properties are selected in any order, the values are always `a` then `d`.  Similarly, selection from array values will always be in the same order.

It gets weird when you have something like `$..*` from

```json
{
  "a": [1, 2, 3],
  "b": [4, 5, 6]
}
```

Here, you must always have 1, 2, and 3 together and in order, as well as 4, 5, and 6 together and in order, but 4, 5, and 6 could come before or after 1, 2, and 3.

glyn:
> Updating my runner wasn't too difficult. All tests pass.

Yes, that's expected since your implementation is more deterministic than the spec.

> 
> It's going to be "fun" filling out the rest of the tests...

Agreed. If we had a full implementation of the RFC which was deliberately non-deterministic in the way it enumerated objects, then we'd have a better chance of automatically discovering tests that should be made non-deterministic (since those tests would fail, at least some of the time).

As it is, there is a risk that we'll miss some tests which should be made non-deterministics. In a sense, that's ok because we can always increase the amount of non-determinism in the CTS, so long as we don't go beyond the spec.

gregsdennis:
> Yes, that's expected since your implementation is more deterministic than the spec.

I might be able to set up the tests in a way to search for objects and scramble them so that the lib pulls the properties out of order.  That would introduce some non-determinism.

It would only be the test runner and wouldn't affect the lib at all.  (I kinda want to play with that now.)

glyn:
> > Yes, that's expected since your implementation is more deterministic than the spec.
> 
> I might be able to set up the tests in a way to search for objects and scramble them so that the lib pulls the properties out of order. That would introduce some non-determinism.
> 
> It would only be the test runner and wouldn't affect the lib at all. (I kinda want to play with that now.)

That would produce _some_ of the non-determinism allowed by the RFC, but not necessarily all of it (e.g. it wouldn't reproduce the non-determinism of the query `$.o[*, *]` shown in [Table 6](https://www.rfc-editor.org/rfc/rfc9535#table-6) of RFC 9535). But it would be a great way of finding some of the tests which need to be made non-deterministic.

glyn:
> For what it's worth, the [nondeterministic](https://github.com/jg-rp/json-p3/tree/nondeterministic) branch of [JSON P3](https://github.com/jg-rp/json-p3) adds some randomness to object iteration.
> 
> After a clone (inc submodules) and `npm install --production=false`, you can run `npm run test query` to test against the main branch of the CTS. Or `npm run test nondeterministic` to test against the nondeterminism branch of glyn/jsonpath-compliance-test-suite.

That's really helpful, thanks for that! (BTW please would you consider adding JSON P3 to the [JSONPath comparison project](https://github.com/cburgmer/json-path-comparison)? A first step is to raise an issue there to add JSON P3.)
> 
> After a few of runs, I've only seen one more case that needs updating, "basic, multiple selectors, wildcard and name", which is less than I was expecting. I might be missing something.

I wasn't expecting large numbers of the current tests to exhibit non-determinism as it requires a combination of factors:

* At least one object (with two or more members) in the query argument and at least one of the following:
1. A wildcard (`*`) in the query applied to one of those objects
2. A filter in the query applied to one of those objects which selects _two or more_ values
3. A descendant segment applied to one of those objects or to one of those object's ancestors

(We could use these criteria to design further non-deterministic tests to plug any gaps.)

gregsdennis:
@jg-rp thanks!  I'll go ahead and merge this, then update that test in another PR.

gregsdennis:
> (I kinda want to play with that now.) - me

Yeah, I tried shuffling objects and it wasn't precise enough.  I get false positives where it thinks a test that is deterministic shouldn't be.

hiltontj:
@glyn - thanks for this. I updated `serde_json_path`'s compliance integration test to handle the non-determinism tests, and its all ✅ now: https://github.com/hiltontj/serde_json_path/pull/85

PR #63, opened 2024-03-03
Add test for filtering object data

jg-rp:
Add a test case for filtering object data, with nondeterministic ordering.


gregsdennis (approved):
This is a good test!  Thanks.  (also works on my lib)

PR #64, opened 2024-03-05
Test that `&&` binds more tightly than `||`

jg-rp:
Test that `&&` binds more tightly than `||`.

`$[?@.a || @.b && @.b]` is testing that `true || false && false` is equivalent to `(true || (false && false))`, which should evaluate to `true`.

`$[?@.b && @.b || @.a]` is testing that `false && false || true` is equivalent to `((false && false ) || true)`, which should evaluate to `true`. This should catch implementations that are inadvertently evaluating from right to left with equal precedence.

I had initially wanted to do `$[?true || false && false]`, but realised literal `true` and `false` are not a "basic-expr". So went with existence tests instead of `$[?true == true || false == false && false == false]`.

### DIFF
--- a/tests/filter.json
+++ b/tests/filter.json
@@ -799,6 +799,44 @@
         [1, 2],
         [2, 1]
       ]
+    },
+    {
+      "name": "and binds more tightly than or",
+      "selector" : "$[?@.a || @.b && @.b]",
+      "document" : [{"a": 1}],
+      "result": [
+        {"a": 1}
+      ]
+    },
+    {
+      "name": "left to right evaluation",
+      "selector" : "$[?@.b && @.b || @.a]",
+      "document" : [{"a": 1}],
+      "result": [
+        {"a": 1}
+      ]
+    },
+    {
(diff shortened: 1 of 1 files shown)


gregsdennis:
I feel we should go ahead and add the grouping tests as well.

- `true && (false || true)`
- `(true || false) && true`

Both are true.

jg-rp:
@gregsdennis I've added those two cases, plus `(true || false) && false`.

Note that `true && false || true` and `true || false && true` will evaluate to `true` with or without parentheses, no matter where they are. Whereas the parens in `(true || false) && false` do change the order of evaluation.

f3ath (approved):

gregsdennis:
> Whereas the parens in `(true || false) && false` do change the order of evaluation.

Yeah, I originally had it this way and decided to change it for some reason.  Both is good.

gregsdennis (approved):
This looks good to me.  Found an issue in my parser as well.

PR #65, opened 2024-03-11
Cover existence tests in logical expressions

jg-rp:
Test that singular filter queries are evaluated as existence tests when they are part of a logical expression, but not involving a comparison expression.

This was [a bug](https://github.com/jg-rp/python-jsonpath/issues/57) for me. I was eagerly unwrapping singular node lists before evaluating comparison expression **and** logical expressions.

### DIFF
--- a/tests/filter.json
+++ b/tests/filter.json
@@ -408,6 +408,23 @@
       ]
     },
     {
+      "name": "exists and exists, data false",
+      "selector" : "$[?@.a&&@.b]",
+      "document" : [{"a": false, "b": false}, {"b": false}, {"c": false}],
+      "result": [
+        {"a": false, "b": false}
+      ]
+    },
+    {
+      "name": "exists or exists, data false",
+      "selector" : "$[?@.a||@.b]",
+      "document" : [{"a": false, "b": false}, {"b": false}, {"c": false}],
+      "result": [
+        {"a": false, "b": false},
+        {"b": false}
+      ]
+    },
+    {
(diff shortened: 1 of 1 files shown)


tests/filter.json
```diff
@@ -407,6 +407,23 @@
         {"a":"c", "d": "f"}
       ]
     },
+    {
+      "name": "exists and exists, data false",
+      "selector" : "$[?@.a&&@.b]",
+      "document" : [{"a": false, "b": false}],
```
gregsdennis, line 413: Can we add an item here that doesn't match?

tests/filter.json
```diff
@@ -407,6 +407,23 @@
...
+      "result": [
+        {"a": false, "b": false}
+      ]
+    },
+    {
+      "name": "exists or exists, data false",
+      "selector" : "$[?@.a||@.b]",
+      "document" : [{"a": false}, {"b": false}],
```
gregsdennis, line 421: Same here?

gregsdennis (changes requested):

tests/filter.json
jg-rp: This test case has been updated to include non-matching data.

tests/filter.json
jg-rp: Non-matching data has been added to this test case, plus data that matches one side of the _or_ expression.

gregsdennis:
Looks good.  Thanks.

PR #73, opened 2024-03-23
Add filter tests for valid non-singular segment

Marcono1234:
Unless I have overlooked something, the existing tests (code in the lines below my changes) was only covering invalid non-singular filters.

These tests also make sure that for example
- `?@[0, 0, 'a']` does not cause the result to contain duplicate values (despite the duplicate `0`)
- a slice such as `?@[0:2]` does not truncate the result if the array has more elements

### DIFF
--- a/tests/filter.json
+++ b/tests/filter.json
@@ -467,6 +467,45 @@
       ]
     },
     {
+      "name": "non-singular existence, wildcard",
+      "selector" : "$[?@.*]",
+      "document" : [1, [], [2], {}, {"a": 3}],
+      "result": [
+        [2],
+        {"a": 3}
+      ]
+    },
+    {
+      "name": "non-singular existence, multiple",
+      "selector" : "$[?@[0, 0, 'a']]",
+      "document" : [1, [], [2], [2, 3], {"a": 3}, {"b": 4}, {"a": 3, "b": 4}],
+      "result": [
+        [2],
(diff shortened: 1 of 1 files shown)


tests/filter.json
```diff
@@ -466,6 +466,45 @@
...
+      "document" : [1, [], [2], {}, {"a": 3}],
+      "result": [
+        [2],
+        {"a": 3}
+      ]
+    },
+    {
+      "name": "non-singluar existence, multiple",
+      "selector" : "$[?@[0, 0, 'a']]",
+      "document" : [1, [], [2], [2, 3], {"a": 3}, {"a": 3, "b": 4}],
```
gregsdennis, line 481: This could use an object that doesn't include an `a` property.  (Would not appear in the result set.)

gregsdennis (approved):
These pass for me.  Just a minor update on one of them.

tests/filter.json
```diff
@@ -466,6 +466,45 @@
         {"d": "f"}
       ]
     },
+    {
+      "name": "non-singluar existence, wildcard",
```
jg-rp, line 470: Typo. "singular" rather than "singluar". Same goes for the following three test cases too.

Marcono1234:
Have corrected the typo in the test names and squashed all changes into a single commit again.

PR #77, opened 2024-05-10
Test I-Regexp mapping corner cases

jg-rp:
This PR covers some corner cases that might arise when mapping I-Regexp patterns to implementation-specific regex patterns.

### DIFF
--- a/tests/functions/match.json
+++ b/tests/functions/match.json
@@ -117,6 +117,48 @@
           "a": "ab"
         }
       ]
+    },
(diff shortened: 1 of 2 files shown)


gregsdennis:
Thanks for these. I'll run them through my implementation next week.

f3ath (approved):
Passes on my implementation

gregsdennis (approved):
Found a bug for me. 👍

PR #78, opened 2024-05-13
add quote escape tests for filter expressions

gregsdennis:
Addresses https://github.com/gregsdennis/json-everything/issues/728

### DIFF
--- a/tests/filter.json
+++ b/tests/filter.json
@@ -932,6 +932,30 @@
       "selector" : "$[?(@.a || @.b) && @.b]",
       "document" : [{"a": 1}],
       "result": []
+    },
+    {
+      "name": "string literal, single quote in double quotes",
+      "selector" : "$[?@ == \"quoted' literal\"]",
+      "document" : [{"a": 1}],
+      "result": []
+    },
+    {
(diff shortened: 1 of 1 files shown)


f3ath (commented):
Pass on my implementation. Any reason why these tests are negative? Seems like adding a positive match would also check that the escaped chars are not missed/mishandled.

gregsdennis:
> Any reason why these tests are negative?

Nope.  Just forgot to update the data.  Will do in a bit (probably tomorrow).

tests/filter.json
```diff
@@ -932,6 +932,30 @@
...
+      "selector" : "$[?@ == 'quoted\\' literal']",
+      "document" : [{"a": 1}],
+      "result": []
+    },
+    {
+      "name": "string literal, escaped double quote in double quotes",
+      "selector" : "$[?@ == \"quoted\\\" literal\"]",
+      "document" : [{"a": 1}],
+      "result": []
     }
```
He-Pin, line 959: How about add tests for named selector too

gregsdennis: Those already exist.

f3ath:
Looks good!

PR #80, opened 2024-05-17
Test that filter expression literals must be compared

jg-rp:
Test that filter expression literals must be compared, unless they're used as arguments to a function expression.


gregsdennis:
@jg-rp why'd you close this?  It looks good to me.

PR #81, opened 2024-05-17
Test that filter expression literals must be compared

jg-rp:
Test that filter expression literals must be compared, unless they're inside a function expression.


gregsdennis (approved):
This looks good by eye. I'll run it through my implementation tomorrow.

f3ath (approved):
LGTM and passes on my implementation

gregsdennis:
Somehow `$[?2]` and `$[?'abc']` don't fail for me.  But they should.

PR #83, opened 2024-06-17
improve boolean operator order of ops and grouping tests

gregsdennis:
Found a few bugs in my implementation surrounding these, and discovered that the tests could be better.


f3ath (approved):
Looks good and passes on my impl.

glyn (commented):
Looks good to me.

PR #84, opened 2024-06-23
Test comparison of empty node lists

jg-rp:
Test that an empty node list is equal to another empty node list and the special result `Nothing`.


f3ath (approved):
LGTM, also passes on my implementation.

PR #86, opened 2024-07-31
Add more tests for invalid escape sequences

Marcono1234:
Covers escape sequences which are allowed in some programming languages (e.g. [C++](https://en.cppreference.com/w/cpp/language/escape) and [Python](https://docs.python.org/3/reference/lexical_analysis.html#escape-sequences)), but which are not valid for JSONPath.

Also adds two test for incomplete `\u` escapes.

I only added these tests for double quoted selectors, but in theory they also apply to single quoted strings (and filters?). But I am not sure if it is really worth duplicating them.

### DIFF
--- a/tests/name_selector.json
+++ b/tests/name_selector.json
@@ -335,6 +335,11 @@
       "invalid_selector": true
     },
     {
+      "name": "double quotes, escape at end of line",
+      "selector": "$[\"\\\n\"]",
+      "invalid_selector": true
+    },
+    {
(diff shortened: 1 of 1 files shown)


gregsdennis (approved):
These look good to me. Thanks for submitting them.

My implementation had a bug in that, while it wasn't accepting these, it was throwing an exception instead of merely saying the path is bad.  (Wrong error issue.)

f3ath (approved):
Pass on my implementation.

PR #87, opened 2024-08-01
Add tests for invalid surrogate pairs

Marcono1234:
Adds tests for invalid surrogate pairs, and one test for directly using a character > U+FFFF (`𝄞`) in the selector, instead of escaping it.

----
Note: Another interesting test would be
- escaped + non-escaped surrogate character, for example: `"selector": "$[\"\\uD800\uDC00\"]"`
(for this a JSONPath implementation might erroneously first convert `\\uD800` to U+D800 before performing validation, and then consider this a valid surrogate pair)
- malformed non-escaped surrogate pairs, for example `"selector": "$[\"\uD800\uD800\"]"`

Note in both cases the single `\` instead of `\\`.

Both are not permitted by JSONPath because it says the query must be valid UTF-8, which is not the case here.

However, it depends on the programming language whether such malformed surrogate pairs are possible. For example for Java whose strings are UTF-16 based such selectors would be valid strings, but the JSONPath implementation would have to reject them.

The problem is that for other programming languages such strings are not valid (e.g. Rust uses UTF-8 for strings), so including for example `"selector": "$[\"\\uD800\uDC00\"]"` might prevent a Rust library from loading `cts.json` at all. Therefore such tests can probably not be added.

### DIFF
--- a/tests/name_selector.json
+++ b/tests/name_selector.json
@@ -200,6 +200,16 @@
       ]
     },
     {
+      "name": "double quotes, supplementary plane character",
+      "selector": "$[\"𝄞\"]",
+      "document": {
+        "𝄞": "A"
+      },
+      "result": [
+        "A"
+      ]
+    },
+    {
       "name": "double quotes, escaped double quote",
       "selector": "$[\"\\\"\"]",
       "document": {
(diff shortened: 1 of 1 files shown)


gregsdennis (approved):
These all fail for me when they're supposed to (but again, they fail incorrectly 😆).

I'm happy with this.  Keep up the good work.

f3ath (approved):
Great job indeed. My implementation was not checking for child selector correctness.

PR #89, opened 2024-08-02
Add tests for invalid and uncommon filter literal values

Marcono1234:
Generally for number literals since JSONPath uses the same syntax as JSON, test values from regular JSON libraries or JSON test suites could be used, for example https://github.com/nst/JSONTestSuite.

So if you have anything else in mind which should be covered, please let me know or feel free to add it yourself.

### DIFF
--- a/tests/filter.json
+++ b/tests/filter.json
@@ -846,7 +846,31 @@
       ]
     },
     {
-      "name": "equals number, exponent leading 0",
+      "name": "equals number, exponent 0",
+      "selector" : "$[?@.a==1e0]",
+      "document" : [{"a": 1, "d": "e"}, {"a":2, "d": "f"}, {"a":"1", "d": "g"}],
+      "result": [
+        {"a": 1, "d": "e"}
+      ]
+    },
+    {
+      "name": "equals number, exponent -0",
+      "selector" : "$[?@.a==1e-0]",
+      "document" : [{"a": 1, "d": "e"}, {"a":2, "d": "f"}, {"a":"1", "d": "g"}],
+      "result": [
+        {"a": 1, "d": "e"}
+      ]
+    },
+    {
+      "name": "equals number, exponent +0",
+      "selector" : "$[?@.a==1e+0]",
+      "document" : [{"a": 1, "d": "e"}, {"a":2, "d": "f"}, {"a":"1", "d": "g"}],
+      "result": [
+        {"a": 1, "d": "e"}
+      ]
+    },
+    {
+      "name": "equals number, exponent leading -0",
       "selector" : "$[?@.a==1e-02]",
       "document" : [{"a": 0.01, "d": "e"}, {"a":0.02, "d": "f"}, {"a":"0.01", "d": "g"}],
       "result": [
@@ -854,7 +878,7 @@
       ]
     },
     {
-      "name": "equals number, exponent 00",
+      "name": "equals number, exponent +00",
       "selector" : "$[?@.a==1e+00]",
       "document" : [{"a": 1, "d": "e"}, {"a":2, "d": "f"}, {"a":"1", "d": "g"}],
       "result": [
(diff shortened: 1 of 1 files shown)


tests/filter.json
```diff
@@ -866,6 +901,91 @@
...
+      "invalid_selector": true
+    },
+    {
+      "name": "equals number, invalid minus space",
+      "selector" : "$[?@.a==- 1]",
+      "invalid_selector": true
+    },
+    {
+      "name": "equals number, invalid double minus",
+      "selector" : "$[?@.a==--1]",
```
gregsdennis, line 916: This one could be interesting if/when we add math operators.

tests/filter.json
```diff
@@ -829,6 +845,22 @@
...
+      "name": "equals number, exponent leading 0",
+      "selector" : "$[?@.a==1e-02]",
+      "document" : [{"a": 0.01, "d": "e"}, {"a":0.02, "d": "f"}, {"a":"0.01", "d": "g"}],
+      "result": [
+        {"a": 0.01, "d": "e"}
+      ]
+    },
+    {
+      "name": "equals number, exponent 00",
+      "selector" : "$[?@.a==1e+00]",
```
gregsdennis, line 858: Do we have a single-zero exponent test?  (1e0, 1e+0, 1e+00)

Or negative zero exponent? (1e-0)

gregsdennis (commented):
Looks good. I think it needs a few more cases. 

Thanks for the work.

tests/filter.json
```diff
@@ -813,6 +821,14 @@
         {"a": 100, "d": "e"}
       ]
     },
+    {
+      "name": "equals number, exponent upper e",
+      "selector" : "$[?@.a==1E2]",
```
gregsdennis, line 826: This isn't valid according to the ABNF

```abnf
number              = (int / "-0") [ frac ] [ exp ] ; decimal number
frac                = "." 1*DIGIT                  ; decimal fraction
exp                 = "e" [ "-" / "+" ] 1*DIGIT    ; decimal exponent
true                = %x74.72.75.65                ; true
false               = %x66.61.6c.73.65             ; false
null                = %x6e.75.6c.6c                ; null
```

Upper `E` is not listed as an allowed exponent character.

gregsdennis (changes requested):
Just some remarks.

tests/filter.json
Marcono1234: `E` seems to be valid (same also for JSON), see https://www.rfc-editor.org/rfc/rfc9535#section-2.3.5.1-13:
> Note: Alphabetic characters in quoted strings are case-insensitive in ABNF, so within a floating point number, the ABNF expression "e" can be either the character 'e' or 'E'.

tests/filter.json
Marcono1234: Good point; but for now should this be kept? If necessary it can be removed / adjusted later in case this becomes valid.

tests/filter.json
Marcono1234: I don't think there were any existing tests; I have added some now.

tests/filter.json
f3ath: IMO it's a good test.

f3ath (approved):
Very nice. Revealed a few issues in my implementation.

tests/filter.json
gregsdennis: Yes, just a comment.

gregsdennis (approved):

He-Pin:
Why treat `+1` as an invalidate , I think we can just create it as valid.

Marcono1234:
> Why treat `+1` as an invalidate , I think we can just create it as valid.

Similar to https://github.com/jsonpath-standard/jsonpath-compliance-test-suite/pull/90#discussion_r1885558117, because that is how it is defined in the JSONPath standard: A leading `+` is not allowed for number literals in filter selectors, see <https://www.rfc-editor.org/rfc/rfc9535#section-2.3.5.1-10>.

(This is also consistent with how JSON numbers are defined in the [JSON specification](https://www.rfc-editor.org/rfc/rfc8259#section-6).)

PR #90, opened 2024-08-02
Add more tests for selectors

Marcono1234:
The "min exact" and "max exact" values here are based on the "range of exact integer values defined in Internet JSON (I-JSON)", as required by the JSONPath specification for index and slice selectors.
I have kept the existing overflow tests because they might still be useful to test overflow during parsing, which might happen before the "range of exact integer values" check in JSONPath implementations.

To be safe, could you please check how your JSONPath implementations behave, to make sure I did not overlook something.

### DIFF
--- a/tests/basic.json
+++ b/tests/basic.json
@@ -171,6 +171,16 @@
       "invalid_selector": true
     },
     {
+      "name": "selector, leading comma",
+      "selector": "$[,0]",
+      "invalid_selector": true
+    },
+    {
+      "name": "selector, trailing comma",
+      "selector": "$[0,]",
+      "invalid_selector": true
+    },
+    {
       "name": "multiple selectors, name and index, array data",
       "selector": "$['a',1]",
       "document": [
@@ -192,6 +202,17 @@
     {
       "name": "multiple selectors, name and index, object data",
       "selector": "$['a',1]",
+      "document": {
+        "a": 1,
+        "b": 2
+      },
+      "result": [
+        1
+      ]
+    },
+    {
+      "name": "multiple selectors, spaces",
+      "selector": "$[ 'a' , 1 ]",
       "document": {
         "a": 1,
         "b": 2
--- a/tests/basic.json
+++ b/tests/basic.json
@@ -211,17 +211,6 @@
       ]
     },
     {
-      "name": "multiple selectors, spaces",
-      "selector": "$[ 'a' , 1 ]",
-      "document": {
-        "a": 1,
-        "b": 2
-      },
-      "result": [
-        1
-      ]
-    },
-    {
       "name": "multiple selectors, index and slice",
       "selector": "$[1,5:7]",
       "document": [
--- a/tests/index_selector.json
+++ b/tests/index_selector.json
@@ -32,6 +32,34 @@
       "result": []
     },
     {
+      "name": "min exact index",
+      "selector": "$[-9007199254740991]",
+      "document": [
+        "first",
+        "second"
+      ],
+      "result": []
+    },
+    {
+      "name": "max exact index",
+      "selector": "$[9007199254740991]",
+      "document": [
+        "first",
+        "second"
+      ],
+      "result": []
+    },
+    {
+      "name": "min exact index - 1",
+      "selector": "$[-9007199254740992]",
+      "invalid_selector": true
+    },
+    {
+      "name": "max exact index + 1",
+      "selector": "$[9007199254740992]",
+      "invalid_selector": true
+    },
+    {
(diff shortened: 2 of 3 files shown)


tests/basic.json
```diff
@@ -200,6 +210,17 @@
         1
       ]
     },
+    {
+      "name": "multiple selectors, spaces",
```
gregsdennis, line 214: This should already be checked in _/tests/whitespace/selectors.json_.  it's probably two separate tests, one for each space.

tests/index_selector.json
```diff
@@ -85,6 +113,26 @@
...
+      "invalid_selector": true
+    },
+    {
+      "name": "plus",
+      "selector": "$[+1]",
+      "invalid_selector": true
+    },
+    {
+      "name": "minus space",
+      "selector": "$[- 1]",
```
gregsdennis, line 128: I think an argument could be made for this to be put in the whitespace tests as well, but I can see it staying here. 

Other opinions welcome.

tests/slice_selector.json
```diff
@@ -41,6 +41,27 @@
         5
       ]
     },
+    {
+      "name": "slice selector with spaces",
```
gregsdennis, line 45: Should already be covered by whitespace tests.

tests/slice_selector.json
```diff
@@ -639,6 +682,36 @@
...
+      "invalid_selector": true
+    },
+    {
+      "name": "plus",
+      "selector": "$[:+1:]",
+      "invalid_selector": true
+    },
+    {
+      "name": "minus space",
+      "selector": "$[:- 1:]",
```
gregsdennis, line 703: Same as with the index variant of this test.

gregsdennis (changes requested):

tests/basic.json
Marcono1234: You are right, sorry I had overlooked those tests.

tests/index_selector.json
f3ath: I'm getting more and more inclined to think we can benefit from tags. I might be able to make an attempt to add tags in a day or two.

tests/slice_selector.json
```diff
@@ -610,6 +610,28 @@
         9
       ]
     },
+    {
```
f3ath, line 613: Would it make sense to add similar checks for the "start" and "step" parameters?

f3ath (approved):
Looks good, let's expand the int boundary tests.

tests/slice_selector.json
gregsdennis: Agreed.  Let's go ahead and add those tests, then this should be good.

Marcono1234: I have duplicated them now for start and step, please let me know if it is ok like this.

tests/index_selector.json
gregsdennis: Yeah, I think a lot of these in particular could be categorized as just "parsing".  I don't think that it's practical to have 2^53 items in a JSON file 😆

gregsdennis (approved):

gregsdennis:
Thanks again @Marcono1234!

tests/index_selector.json
```diff
@@ -85,6 +113,26 @@
...
       "invalid_selector": true
     },
+    {
+      "name": "decimal",
+      "selector": "$[1.0]",
+      "invalid_selector": true
+    },
+    {
+      "name": "plus",
+      "selector": "$[+1]",
```
He-Pin, line 123: Why is this invalid ? in Java, this is valid.
```java
  private def `int`[_: P]: P[Int] = P("0" | (CharIn("+\\-").? ~ `DIGIT1` ~ `DIGIT`.rep)).!
    .map(_.toLong)
    .filter(value => MIN_INTEGER < value && value < MAX_INTEGER)
    .map(num => Math.min(Int.MaxValue, Math.max(Int.MinValue, num)).toInt)
```

f3ath: The standard does not allow `+` for integers.
```
index-selector      = int                        ; decimal integer

int                 = "0" /
                      (["-"] DIGIT1 *DIGIT)      ; - optional
DIGIT1              = %x31-39                    ; 1-9 non-zero digit
```
https://ietf-wg-jsonpath.github.io/draft-ietf-jsonpath-base/draft-ietf-jsonpath-base.html#name-syntax-5

He-Pin: Oh, I see. It seems I missed this, but I think `+` can also be optional, will update my implementation, thanks

PR #91, opened 2024-08-03
Add more surrogate pair tests

Marcono1234:


f3ath (approved):
Looks good and works on my end.

Issue #36, opened 2023-04-26
Introduce test ID


f3ath:
Some implementations may want to ignore specific tests, e.g. due to hard limitation of the underlying platforms. For such cases we might want to introduce a test ID. Such an ID should be stable, i.e. it should not change when the test metadata (like name) changes, neither it should depend on the order of the results.

I am considering adding the `id` field to the test object:
```json
    {
      "name": "name shorthand",
      "id": "17e682f060b5f8e47ea04c5c4855908b0a5ad612022260fe50e11ecb0cc0ab76",
      "selector": "$.a",
      "document": {
        "a": "A",
        "b": "B"
      },
      "result": [
        "A"
      ]
```

The `id` is calculated as follows:
1. take the `test` object
2. remove the `name` key
3. sort the `result` array if it exists (this should not affect the semantics since the traversal order is explicitly undefined)
4. apply JSON encode
5. take `sha256` hash sum of the encoded string

@gregsdennis @glyn let me know what you think.

gregsdennis:
I just use the test name.

https://github.com/gregsdennis/json-everything/blob/master/JsonPath.Tests/Suite/ComplianceTestSuiteTests.cs#L15-L17

I don't think we have anything in the CI that guarantees uniqueness, though.  We'd have to put something in whether we use the name or an id.

glyn:
My general view of this change is that it's going to make maintenance more difficult. Better to simply ensure that all the test case names are unique - something that could be checked when cts.json is generated.
 
>     3. sort the `result` array if it exists (this should not affect the semantics since the traversal order is explicitly undefined)

The order of the result array is partially defined by the spec. For instance [*] applied to an array produces the elements of the array in array order.

That said we still need to address issue #1, which may further complicate matters.

f3ath:
This is partially solved by #92. I don't see much value in keeping this open.

PR #93, opened 2024-08-04
Check for duplicate test names

Marcono1234:
Inspired by https://github.com/jsonpath-standard/jsonpath-compliance-test-suite/issues/36#issuecomment-1523515817

Due to copy & paste errors there were already duplicate names in `selectors.json`, see changes below.

Not sure if the change in `build.js` clean enough, please let me know if I should change it in some way.

### DIFF
--- a/build.js
+++ b/build.js
@@ -62,13 +62,16 @@ const usage = [
     '  -e, --exclude        Flag to exclude the specified tags instead of including them.',
 ]
 
+const exitCodeUsageError = 1;
+const exitCodeDuplicateNames = 2;
+
 main()
 
 function main() {
     const {values, positionals} = parseArgs({options, allowPositionals: true});
     if (positionals.length !== 1) {
         usage.forEach(line => console.error(line));
-        process.exit(1);
+        process.exit(exitCodeUsageError);
     }
     const [testsFolder] = positionals;
     buildTestSuite(testsFolder, values['tag'], values['require-all'], values['exclude']);
@@ -90,8 +93,10 @@ function buildTestSuite(testsFolder, tags, all, exclude) {
     const tagFilter = all
         ? list => tags.every(t => list.includes(t))
         : list => tags.some(t => list.includes(t))
-    const tests = readTestsFromDir(testsFolder)
-        .filter(test => tags.length === 0 || (tagFilter(getTags(test)) !== exclude));
+    let tests = readTestsFromDir(testsFolder);
+    checkDuplicateNames(tests);
+
+    tests = tests.filter(test => tags.length === 0 || (tagFilter(getTags(test)) !== exclude));
     tests.forEach(test => {
         if ('tags' in test) test.tags.sort();
     });
@@ -194,3 +199,27 @@ function prependToTestName(prefix) {
     };
 }
 
+/**
+ * Verifies that no tests have duplicate names.
+ * @param allTests - The list of all tests.
+ */
+function checkDuplicateNames(allTests) {
+    const names = new Set();
+    const duplicates = [];
+
+    for (const test of allTests) {
+        const name = test.name;
+
+        if (names.has(name)) {
+            duplicates.push(name);
+        } else {
+            names.add(name);
+        }
+    }
+
+    if (duplicates.length > 0) {
+        duplicates.sort();
+        console.error(['Error: Duplicate test names:', ...duplicates].join('\n '));
+        process.exit(exitCodeDuplicateNames);
+    }
+}
--- a/tests/whitespace/selectors.json
+++ b/tests/whitespace/selectors.json
@@ -36,21 +36,21 @@
       "tags": ["whitespace"]
     },
     {
-      "name": "newline between root and bracket",
+      "name": "newline between bracket and bracket",
       "selector" : "$['a'] \n['b']",
       "document" : {"a": {"b": "ab"} },
       "result": [ "ab" ],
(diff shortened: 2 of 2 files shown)


gregsdennis:
Can you split the changes so that we can see the build fail?  Put the name check on the first commit, and let it build.  Then come back and fix the test names, which should let the build pass.

build.js
```diff
@@ -194,3 +196,29 @@ function prependToTestName(prefix) {
...
+    const names = new Set();
+    const duplicates = [];
+
+    for (const test of allTests) {
+        const oldSize = names.size;
+        const name = test.name;
+        names.add(name);
+
+        // If size remained the same, a duplicate was encountered
+        if (oldSize === names.size) {
```
f3ath, line 213: You can get rid of `oldSize` and do `if (names.has(test.name)) { ... }` instead.

build.js
```diff
@@ -194,3 +196,29 @@ function prependToTestName(prefix) {
...
+            duplicates.push(name);
+        }
+    }
+
+    if (duplicates.length > 0) {
+        duplicates.sort();
+        const prefix = "\n  ";
+        console.error("Error: Duplicate test names:" + prefix + duplicates.join(prefix));
+        process.exit(1);
```
f3ath, line 222: Do we really want to fail the build or just emit a warning? Not suggesting, just asking.

build.js
```diff
@@ -194,3 +196,29 @@ function prependToTestName(prefix) {
...
+
+        // If size remained the same, a duplicate was encountered
+        if (oldSize === names.size) {
+            duplicates.push(name);
+        }
+    }
+
+    if (duplicates.length > 0) {
+        duplicates.sort();
+        const prefix = "\n  ";
```
f3ath, line 220: nit: Seems like we're using single quotes everywhere else.

build.js
```diff
@@ -194,3 +196,29 @@ function prependToTestName(prefix) {
...
+        console.error("Error: Duplicate test names:" + prefix + duplicates.join(prefix));
```
f3ath, line 221: nit: `console.error(['Error: Duplicate test names:', ...duplicates].join('\n '));`

f3ath, line 221:
```suggestion
        cconsole.error(['Error: Duplicate test names:', ...duplicates].join('\n '));
```

build.js
```diff
@@ -92,6 +92,8 @@ function buildTestSuite(testsFolder, tags, all, exclude) {
         : list => tags.some(t => list.includes(t))
     const tests = readTestsFromDir(testsFolder)
         .filter(test => tags.length === 0 || (tagFilter(getTags(test)) !== exclude));
+    checkDuplicateNames(tests);
```
f3ath, line 95: We might want to check for dupes before applying `.filter(...)`.

build.js
f3ath: If we do fail, it makes sense to pick a different exit code and define them at the top of the script.

f3ath:
The build script may soon require its own test suite...

build.js
Marcono1234: At least there should be some flag then to let CI fail for duplicate test names. But I think it might be best to always fail. Duplicate names will just be confusing.

gregsdennis: GH Actions doesn't support warnings (as far as I know).  At least, they're not visible on this "conversation" page.  You have to go into the file changes to see anything.  It shows up like this:

image

I think I'd prefer the failure.

build.js
Marcono1234: I was trying to make this a bit more efficient (hoping that `size` is not recalculated every time), because `has` + `add` would mean there are two 'contains' checks internally. It is quite unfortunate that `add` does not return whether it was successful.

But maybe that was premature optimization.

build.js
Marcono1234: Is it ok then if I change `const tests` to `let tests`, and reassign it after the check with the filtered tests?

build.js
f3ath: I believe that however inefficient it is complexity-wise, it'll still be negligible compared to the IO.

gregsdennis: (I also tend to care less about efficiency on a build server.  Just my opinion, though.)

build.js
f3ath: Rr `allTests` and `filteredTests`.

build.js
f3ath: I'm fine either way, just let's use another exit code, `1` is used here already https://github.com/jsonpath-standard/jsonpath-compliance-test-suite/blob/733fe526ceccb183ed0e3397340f2ee9feec0d67/build.js#L71

gregsdennis: @f3ath I'm curious of your reason why.  Does it make a different in local development?

f3ath: Just future-proofing in case we introduce more checks and/or want to act differently depending on why it failed.

Marcono1234:
Have split it into two commits now: The build for the first commit [failed](https://github.com/jsonpath-standard/jsonpath-compliance-test-suite/actions/runs/10272891697/job/28426122167?pr=93#step:4:26), as desired. And the second commit fixed it.

I hope the changes are ok like this now. Please let me know if you want anything else changed.

gregsdennis:
Hmm... @Marcono1234 / @f3ath I accidently merged this. Meant to approve instead and wait for @f3ath's responses.  Please open a new PR for any additional changes.

Issue #72, opened 2024-03-23
Cover absolute path filter queries


Marcono1234:
It might be good to add tests for using an absolute path (that is `$` instead of `@`) in a filter selector, e.g. `$[?$.a==1]`. Currently the filter tests only seem to cover relative paths (`@`) or invalid selectors with absolute paths, but not any valid selectors with absolute paths.

Might be good to cover:
- existence test without segment: `$[?$]`
- some 'normal' filter expressions, e.g. `$[?$.a==1]`
- using `(...)`, e.g. `$[?($.a==1)]`
- applying the filter selector to a nested value and making sure `$` is evaluated absolute and not relative, for example:
  - selector: `$.a[?$.b==2]`
  - document: `{"a": [{"b": 1}], "b": 2}`
  
What do you think?

gregsdennis:
We have quite a few of them actually.  See tests with "absolute non-singular query" in the title.

https://github.com/jsonpath-standard/jsonpath-compliance-test-suite/blob/7bd8532dfcccd288bbcaa67b48e5804911991094/tests/filter.json#L573-L647

glyn:
Sure!

gregsdennis:
Yes, happy to see a PR that fills any gaps.

PR #97, opened 2025-01-15
Add absolute path filter tests

taimoorzaeem:
I think these are sufficient for my test coverage for now. I might add more in the future if need be.

Related #72.

### DIFF
--- a/tests/filter.json
+++ b/tests/filter.json
@@ -26,6 +26,24 @@
       ]
     },
     {
+      "name": "absolute existence, without segments",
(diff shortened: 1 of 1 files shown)


gregsdennis (approved):
These pass for me.  I'm not sure what the value of them is, but they're valid per the spec, so I suppose they should be included.

gregsdennis:
@f3ath please feel free to merge once you've had a look.

f3ath (approved):
LGTM and works on my end.

Issue #99, opened 2025-01-23
Testing normalized paths


jg-rp:
I find myself needing to test normalized paths (the string representation a node's location according to [section 2.7 of RFC 9535](https://datatracker.ietf.org/doc/html/rfc9535#name-normalized-paths)) in multiple projects.

To that end I've started a CTS-style [JSONPath Normalized Path Test Suite](https://github.com/jg-rp/jsonpath-compliance-normalized-paths). Does such a thing already exist? If not, is this useful elsewhere and is it worth moving it (or something similar) to a jsonpath-standard repo?

 [JSONPath Normalized Path Test Suite](https://github.com/jg-rp/jsonpath-compliance-normalized-paths) includes tests for "canonical paths" too. Although not defined by RFC 9535, I wanted a consistent string representation of a compiled JSONPath query, which is not a normalized path and is not necessarily a singular query.

taimoorzaeem:
@jg-rp I appreciate the initiative. Personally, I would love normalized paths to be included in `cts.json`. Something like below would have been perfect. A new repository seems too decentralized than it has to be.

```js
{
      "name": "basic, root",
      "selector": "$",
      "document": [
        "first",
        "second"
      ],
      "result": [
        [
          "first",
          "second"
        ]
      ],
     "result-located": [
        { "loc": "$",
          "node": [
             first,
             second
          ]
        }
     }
}
```

f3ath:
@taimoorzaeem what is the purpose of the `node` key in your example?

@jg-rp I would imagine adding optional `path` or `paths` keys to the test. The `path` being an array of the normalized paths of the corresponding values in the `result`. The `paths` being an array of arrays of normalized paths corresponding to the values in `results`.

```json
    {
      "name": "filter, existence",
      "selector": "$[?@.a]",
      "document": [
        {
          "a": "b",
          "d": "e"
        },
        {
          "b": "c",
          "d": "f"
        }
      ],
      "result": [
        {
          "a": "b",
          "d": "e"
        }
      ]
    },
    "path": [
        "$[0]"
      ]
    },

```

Perhaps we can allow tests without `result` and `results` if they contain either `path` or `paths`.

taimoorzaeem:
> @taimoorzaeem what is the purpose of the `node` key in your example?

Oh, now that I see, we don't really need that. I like your example better.

f3ath:
[...]

PR #100, opened 2025-01-24
Test normalized paths

jg-rp:
Similar to @f3ath's suggestion in #99, this PR changes the schema to allow `result_paths` or `results_paths` properties (I couldn't quite bring myself to use `path` singular for an array) and adds expected normalized paths to a few select test cases.

Currently, `result_paths` is only allowed when `result` is present, and `results_paths` is only allowed if `results` is present. I imagine if normalized paths are worth testing for a given query then the resulting values should be tested too, so testing paths alone is not currently allowed by the schema.


f3ath (approved):
LGTM. Will wait for approvals from @glyn and @gregsdennis.

hiltontj (commented):
Nice work! This is a great idea, a user had recently raised an issue on `serde_json_path`'s normalized path generation. See https://github.com/hiltontj/serde_json_path/pull/113

cts.schema.json
```diff
@@ -112,7 +122,19 @@
...
       "description": "The flag indicating that the selector is not a valid JSONPath selector expression"
+    },
+    "result_paths": {
+      "description": "The expected normalized JSONPath query strings, one for each result value",
+      "type": "array",
+      "items": {
+        "type": "string"
+      }
+    },
+    "results_paths": {
```
gregsdennis, line 133: Ordinarily, I'd consider this an unnecessary refactor. Since the definition is only used in one place, it makes sense to just inline it. 

But we're doing this for other definitions, so... okay.

gregsdennis (approved):

f3ath:
I just ran the new tests in my implementation, all passed.

cts.schema.json
```diff
@@ -112,7 +122,19 @@
...
+      "description": "An array of possible expected result paths",
```
glyn, line 134: I think that description is a little unclear.
```suggestion
      "description": "An array of possible expected normalized JSONPath query strings of applying the selector to the document, each element of which contains all the normalized JSONPath query strings of matched values",
```

cts.schema.json
```diff
@@ -112,7 +122,19 @@
       "type": "boolean",
       "const": true,
       "description": "The flag indicating that the selector is not a valid JSONPath selector expression"
+    },
+    "result_paths": {
+      "description": "The expected normalized JSONPath query strings, one for each result value",
```
glyn, line 127: I think this could be made clearer:
```suggestion
      "description": "The expected normalized JSONPath query strings of applying the selector to the document, contains all the normalized JSONPath query strings of matched values",
```

glyn (changes requested):
I think this is valuable, thanks. But I don't understand why we'd want some tests _not_ to have normalized paths. Would it be possible to add a commit which includes all the normalized paths for all remaining tests and also have the schema mandate that result_paths is present whenever result is present and results_paths is present whenever results is present?

jg-rp:
> But I don't understand why we'd want some tests not to have normalized paths. Would it be possible to add a commit which includes all the normalized paths ..

It's possible. I was thinking 100% path test coverage would be something we'd work towards over time, over multiple commits, considering the labour involved in manually changing all 679 test cases. I'll happily chip away at them here over the coming days if necessary.

Alternatively we could generate/validate all expected normalized paths programatically using one or more existing implementations, which would give me reasonable confidence in the correctness of the resulting tests, but not absolute confidence.

If we were to generate normalized paths programmatically, it would make an alternative schema seem less daunting. For example:

(Note that I've changed `selector` to `query` in these examples to match RFC 9535 terminology.)

```json
{
  "name": "basic, wildcard shorthand, array data",
  "query": "$.*",
  "document": ["first", "second"],
  "result": {
    "values": ["first", "second"],
    "paths": ["$[0]", "$[1]"]
  }
}
```

and

```json
{
  "name": "basic, wildcard shorthand, object data",
  "query": "$.*",
  "document": {
    "a": "A",
    "b": "B"
  },
  "results": {
    "values": [
      ["A", "B"],
      ["B", "A"]
    ],
    "paths": [
      ["$['a']", "$['b']"],
      ["$['b']", "$['a']"]
    ]
  }
}
```

glyn:
> Alternatively we could generate/validate all expected normalized paths programatically using one or more existing implementations, which would give me reasonable confidence in the correctness of the resulting tests.

Yup, I think that's the best solution. I'd be comfortable going with the paths that two implementations (from distinct implementers) agree on.

gregsdennis:
Mine already outputs node locations. I'll see if I can batch the output.

f3ath:
@glyn @gregsdennis I got my version of normalized paths in #101 . We can compare the compiled cts.json to cross-check.

jg-rp:
My last commit populates `result_paths` or `results_paths` for all test cases.

`cts.json` failing validation is expected as the schema now requires  `result_paths` or `results_paths` for valid cases.

@f3ath, it looks like #101 has not updated files in `tests/functions` and `tests/whitespace` and your "basic, multiple selectors, wildcard and name" seems to have too many items in `results_paths`.

f3ath:
Thanks for the review @jg-rp . I have updated #101.

cts.schema.json
```diff
@@ -112,7 +124,19 @@
...
+      "description": "The expected normalized JSONPath query strings of applying the selector to the document, contains all the normalized JSONPath query strings of matched values",
+      "type": "array",
+      "items": {
+        "type": "string"
+      }
+    },
+    "results_paths": {
+      "description": "An array of possible expected normalized JSONPath query strings of applying the selector to the document, each element of which contains all the normalized JSONPath query strings of matched values",
+      "type": "array",
+      "items": {"$ref": "#/$defs/result_paths"},
```
gregsdennis, line 138: Was this supposed to be a recursive definition?

gregsdennis:
Looks like I have a bug where I'm not properly escaping chars in name selectors, but everything else passes for me.

Also, I realized that I can't fully generate the cases because I only handle a single potential ordering.

cts.schema.json
jg-rp: Hi @greg, I assumed it was done this way so `test_case_result` (and now `test_case_results`) could be referenced remotely by URL form another schema. Happy to change it if you'd prefer it inline.

f3ath:
> Also, I realized that I can't fully generate the cases because I only handle a single potential ordering.

I used the existing `results` values to generate the corresponding `results_values`.

cts.schema.json
gregsdennis: It's fine as is

jg-rp:
@glyn, updated `cts.json` is now included and passes validation.

glyn (approved):
Thanks. Nice work @jg-rp and @f3ath.

gregsdennis:
Just started working on my bug, and I see that it's just something I never came back to. 😆 

image

PR #101, opened 2025-01-26
Generated normalized paths

f3ath:
Generated normalized paths as suggested in https://github.com/jsonpath-standard/jsonpath-compliance-test-suite/pull/100#issuecomment-2614016926


f3ath:
Updated after the review https://github.com/jsonpath-standard/jsonpath-compliance-test-suite/pull/100#issuecomment-2614332154

gregsdennis (approved):
Same as on the other PR: my implementation is having trouble with escaping chars in name selectors.  Otherwise passes.

tests/name_selector.json
```diff
@@ -282,56 +372,74 @@
       },
       "result": [
         "A"
+      ],
+      "result_paths": [
+        "$['\\\\/']"
```
jg-rp, line 377: Should the path be `"$['/']"` here? I think solidus (`/`, 0x2F) is "normal-unescaped".

Same for "single quotes, escaped solidus".

jg-rp:
Comparing `cts.json` from this PR to one built from #100, #100 adds three test cases:

- basic, name shorthand, object data, nested
- name selector, name, double quotes, nested
- name selector, name, double quotes, contains

And differs on the expected normalized path for:

- name selector, double quotes, escaped solidus
- name selector, single quotes, escaped solidus

Otherwise every test case matches perfectly.

tests/name_selector.json
f3ath: Yeah this seems to be a bug in my implementation. The document does not have backslashes, so why would they appear in the normalized path.

Thanks for catching all these things.

f3ath: On a second thought though, according to 2.3.1.1, the `/` is escapable
```
   name-selector       = string-literal

   string-literal      = %x22 *double-quoted %x22 /     ; "string"
                         %x27 *single-quoted %x27       ; 'string'

   double-quoted       = unescaped /
                         %x27      /                    ; '
                         ESC %x22  /                    ; \"
                         ESC escapable

   single-quoted       = unescaped /
                         %x22      /                    ; "
                         ESC %x27  /                    ; \'
                         ESC escapable

   ESC                 = %x5C                           ; \ backslash

   unescaped           = %x20-21 /                      ; see RFC 8259
                            ; omit 0x22 "
                         %x23-26 /
                            ; omit 0x27 '
                         %x28-5B /
                            ; omit 0x5C \
                         %x5D-D7FF /
                            ; skip surrogate code points
                         %xE000-10FFFF

   escapable           = %x62 / ; b BS backspace U+0008
                         %x66 / ; f FF form feed U+000C
                         %x6E / ; n LF line feed U+000A
                         %x72 / ; r CR carriage return U+000D
                         %x74 / ; t HT horizontal tab U+0009
                         "/"  / ; / slash (solidus) U+002F
                         "\"  / ; \ backslash (reverse solidus) U+005C
                         (%x75 hexchar) ;  uXXXX U+XXXX
```

I'll need to take a deeper look.

jg-rp: > On second thought though, according to 2.3.1.1, the `/` is escapable

True, but [2.7](https://datatracker.ietf.org/doc/html/rfc9535#name-normalized-paths) says 

> Certain characters are escaped in Normalized Paths in one and only one way; all other characters are unescaped.

And the grammar puts `/` in the unescaped category.

```
normal-single-quoted = normal-unescaped /
                       ESC normal-escapable
normal-unescaped     =    ; omit %x0-1F control codes
                       %x20-26 /
                          ; omit 0x27 '
                       %x28-5B /
                          ; omit 0x5C \
                       %x5D-D7FF /
                          ; skip surrogate code points
                       %xE000-10FFFF

normal-escapable     = %x62 / ; b BS backspace U+0008
                       %x66 / ; f FF form feed U+000C
                       %x6E / ; n LF line feed U+000A
                       %x72 / ; r CR carriage return U+000D
                       %x74 / ; t HT horizontal tab U+0009
                       "'" /  ; ' apostrophe U+0027
                       "\" /  ; \ backslash (reverse solidus) U+005C
                       (%x75 normal-hexchar)
                                       ; certain values u00xx U+00XX
```

f3ath: So does this mean that `$['/']` is a correct normalized path (as per [2.7](https://datatracker.ietf.org/doc/html/rfc9535#section-2.7)) but is not a valid JSONPath expression (as per [2.3.1.1](https://datatracker.ietf.org/doc/html/rfc9535#section-2.3.1.1))? @glyn @gregsdennis

f3ath: It seems like it's worth clarifying in the RFC repo: https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/526

f3ath: OK, so this is indeed a bug in my implementation.

glyn: > So does this mean that `$['/']` is a correct normalized path (as per [2.7](https://datatracker.ietf.org/doc/html/rfc9535#section-2.7)) but not a valid JSONPath query (as per [2.3.1.1](https://datatracker.ietf.org/doc/html/rfc9535#section-2.3.1.1))? @glyn @gregsdennis
> 

(I think this question has been answered by https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/526.)

f3ath: Yes. I incorrectly interpreted "escapable" as "must be escaped".

jg-rp:
I have rerun my `cts.json` comparison script. This PR now matches #100, with the exception of the three additional test cases mentioned above and the updates to `cts.schema.json`.

gregsdennis:
@jg-rp @f3ath which PR are we going for, this or #100?

glyn:
Whichever PR we go for, I'd like to see the schema changes included so that
normalized paths are mandatory.

jg-rp:
> @jg-rp @f3ath which PR are we going for, this or #100?

Either works for me. 👍

glyn:
This PR doesn't have the correct schema changes. PR #100 has the correct schema changes, but is missing some JSON and so the schema validation check fails.

jg-rp:
> This PR doesn't have the correct schema changes. PR #100 has the correct schema changes, but is missing some JSON and so the schema validation check fails.

Indeed. Excluding `cts.json` from #100 is deliberate, as it gets built by CI. I’d be happy to include it if you prefer.

I get the impression that #101 is intended to aid comparison and is not to be merged, at least not in its current form.

glyn:
> > This PR doesn't have the correct schema changes. PR #100 has the correct schema changes, but is missing some JSON and so the schema validation check fails.
> 
> Indeed. Excluding an updated `cts.json` from #100 is deliberate, as it gets built by CI. I’d be happy to include it if you prefer.

Yes, please. It should get overwritten with the same content by CI, but at least the schema validation should pass. (Aside: strange that it validates cts.json _before_ overwriting it rather than _after_.)

> 
> I get the impression that #101 is intended to aid comparison and is not to be merged, at least not in its current form.

Ok, let's close 101 to be crystal clear what's going on.

f3ath:
> I get the impression that #101 is intended to aid comparison and is not to be merged, at least not in its current form.

Correct. This PR was to help cross check the paths.

PR #102, opened 2025-02-08
Add tests of selector that follows another selector.  Resolves #96

### LINKED ISSUE #96: Add test for name selector that follows a bracketed selector
I'm at the point where I've just written a jsonpath parser, and it passes all of the tests in the compliance test suite.
Now I'm integrating my implementation into some existing software, replacing a non-standards-compliant jsonpath library that was already being used.
So, I'm finding bugs in my own parser that are not covered by the compliance test suite.

This is the latest one.

The CTS contains no tests where a name selector follows a bracketed selector such as a filter selector.
This is supported, since the RFC contains this example:
```
$.store.book[?@.price < 10].title
```

I don't see any similar examples based on grepping `cts.json`:
```
$ rg '\].*\.\w' cts.json
2236:      "selector": "$[?@.a[*].a==0]",
8162:      "selector": "$..[?length(@)==length($ [0] .a)]",
8180:      "selector": "$..[?length(@)==length($\n[0]\n.a)]",
8198:      "selector": "$..[?length(@)==length($\t[0]\t.a)]",
8216:      "selector": "$..[?length(@)==length($\r[0]\r.a)]",
```

gongfarmer:
Add test slice selector followed by slice selector
Add test filter selector followed by child segment that selects multiple elements
Add test filter selector followed by name selector

### DIFF
--- a/tests/filter.json
+++ b/tests/filter.json
@@ -1925,6 +1925,20 @@
       ]
     },
     {
+      "name": "followed by child segment that selects multiple elements",
+      "selector": "$[?@.z=='_']['x','y']",
+      "document": [
+        {
+          "x": 1, "y": null, "z": "_"
+        }
+      ],
+      "result": [ 1, null ],
+      "result_paths": [
+        "$[0]['x']",
+        "$[0]['y']"
+      ]
+    },
+    {
       "name": "relative non-singular query, index, equal",
       "selector": "$[?(@[0, 0]==42)]",
       "invalid_selector": true,
--- a/tests/filter.json
+++ b/tests/filter.json
@@ -1911,6 +1911,20 @@
       "result_paths": []
     },
     {
+      "name": "followed by name selector",
+      "selector": "$[?@.a==1].b.x",
+      "document": [
+        {
+          "a": 1,
+          "b": { "x": 2 }
+        }
+      ],
+      "result": [ 2 ],
+      "result_paths": [
+        "$[0]['b']['x']"
+      ]
+    },
+    {
       "name": "relative non-singular query, index, equal",
       "selector": "$[?(@[0, 0]==42)]",
       "invalid_selector": true,
--- a/tests/slice_selector.json
+++ b/tests/slice_selector.json
@@ -524,6 +524,55 @@
       ]
     },
     {
+      "name": "in serial, on nested array",
+      "selector": "$[1:3][1:2]",
+      "document": [
+        [
+          "a",
+          "b",
+          "c"
+        ],
+        [
+          "d",
+          "e",
+          "f"
+        ],
+        [
+          "g",
+          "h",
+          "i"
+        ]
+      ],
+      "result": [
+        "e",
+        "h"
+      ],
+      "result_paths": [
+        "$[1][1]",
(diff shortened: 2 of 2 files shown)


tests/filter.json
```diff
@@ -1910,6 +1910,34 @@
...
+      "name": "followed by child segment that selects multiple elements",
+      "selector": "$[?@.z=='_']['x','y']",
+      "document": [
+        {
+          "a": { "x": 1, "y": null, "z": "_" }
+        }
+      ],
+      "result": [ 1, null ],
+      "result_paths": [
+        "$['a']['x']",
```
f3ath, line 1937: This one does not seem right. The root element of the document is an array.

f3ath (changes requested):
The "followed by child segment that selects multiple elements" seems off.

gongfarmer:
> The "followed by child segment that selects multiple elements" seems off.

It's named that way because when cts.json is built, this name is appended onto the name of the file.
So within `cts.json` the test is named  "filter, followed by child segment that selects multiple elements"

Is that the issue?

f3ath:
> > The "followed by child segment that selects multiple elements" seems off.
> 
> It's named that way because when cts.json is built, this name is appended onto the name of the file. So within `cts.json` the test is named "filter, followed by child segment that selects multiple elements"
> 
> Is that the issue?

Sorry, I didn't mean the name itself. See the comment here: https://github.com/jsonpath-standard/jsonpath-compliance-test-suite/pull/102#discussion_r1950201679

tests/filter.json
gongfarmer: `$[0]` would select the first array element, which is the hash.  This is not what is wanted here - the filter selector is supposed to be applied to the array, not the hash. 

For comparison, see the other tests within `filter.json`.
```
     5       "selector": "$[? @.a]",
    31       "selector": "$[?\n@.a]",
    57       "selector": "$[?\t@.a]",
    83       "selector": "$[?\r@.a]",
 ...
 ```
In every test, the document is an array, and the filter selector is applied directly to the root identifer:

Maybe it's confusing that I only put one element in the array, so there's nothing to show what is not being selected by the filter.  Would it help to add another array element?

f3ath: The root element of the document in this test is an array with a single element:
```json
       [
        {
          "a": { "x": 1, "y": null, "z": "_" }
        }
      ]
 ```
 
 therefore anything in the "result_paths" has to start with an array index: `$[0]`, but the result paths provided start with `$['a']` as if the root element was an object. Does the document have extra square brackets?

gongfarmer: Oh I see, you were referring to the `result_paths`, which is exactly what your comment is on.  I should have seen that.

Yes I agree.  I'll correct that and re-test.

gongfarmer: I pushed an updated version.  

I corrected the paths and changed the document for this test, which did not match the selector.
(The { "a": ... } hash should not have been there, only the single value it contains.)

f3ath (approved):
LGTM

f3ath:
Passes on my implementation. @gregsdennis can you please check on your end?

PR #105, opened 2025-04-24
Improve build_cts_json.yaml workflow security

Marcono1234:
Limits the increased permissions to a dedicated job which runs only on push events.

The previous implementation was already enough to protect against malicious PRs from forks, see for example [this run](https://github.com/jsonpath-standard/jsonpath-compliance-test-suite/actions/runs/13380998016/job/37369438034#step:1:17), which only has `Contents: read` as desired.

However, if Dependabot was used on the repository (at some point in the future), it would have increased its default read-only permissions to write permissions, making it easier for compromised dependencies to directly affect the repository.

With the new changes an unprivileged job builds the `cts.json` now, and more privileged job pushes the changes, but only when running on `main`. The `cts.json` file is passed as artifact between the jobs (as described in the [documentation](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/storing-and-sharing-data-from-a-workflow#passing-data-between-jobs-in-a-workflow)).

### DIFF
--- a/.github/workflows/build_cts_json.yaml
+++ b/.github/workflows/build_cts_json.yaml
@@ -5,21 +5,18 @@ on:
       - main
   pull_request:
 
+permissions: {}
+
 jobs:
   build-cts:
     runs-on: ubuntu-latest
-
     permissions:
-      # Allow the job to push the changed file to the repository
-      # For pull requests from forks this seems to be implicitly changed to 'read', as desired, see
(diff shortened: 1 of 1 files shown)


Marcono1234:
If you don't think this is worth it or needed, feel free to close this pull request. As mentioned above, the existing workflow should already be safe against malicious PRs from external forks.

PR #110, opened 2025-09-07
Test JSON bool, null literal shorthand names

theory:
Section 2.5.1.1 defines shorthand names with this grammar:

```anbf
member-name-shorthand = name-first *name-char
name-first          = ALPHA /
                      "_"   /
                      %x80-D7FF /
                         ; skip surrogate code points
                      %xE000-10FFFF
name-char           = name-first / DIGIT
```

This allows shorthand names that correspond to the JSON literals `true`, `false`, and `null`.

So add tests for this pattern.

### DIFF
--- a/tests/basic.json
+++ b/tests/basic.json
@@ -565,6 +565,57 @@
       ]
     },
     {
+      "name": "name shorthand, true",
+      "selector": "$.true",
+      "document": {
+        "true": "A",
+        "_foo": "B"
+      },
+      "result": [
+        "A"
+      ],
+      "result_paths": [
+        "$['true']"
+      ],
+      "tags": [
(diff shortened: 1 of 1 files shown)


taimoorzaeem:
Hi @theory, these tests look good to merge.

I was just wondering, what's the use case for covering these cases? I can't think of any case where an implementation would work fine with `$.name` but not with `$.true`. :thinking:

theory:
Here's one: [mine](https://github.com/theory/jsonpath/pull/22/commits/deb0af2). I had overlooked including bools and null in a switch statement. :-)

f3ath (approved):
LG and passes for me. @gregsdennis any objections?

PR #113, opened 2025-11-21
Test quoted false positive of incorrect capitalization

TamasSzigeti:
This came up in the [Symfony implementation](https://github.com/symfony/symfony/pull/62465), I was told it might be useful upstream too

### DIFF
--- a/tests/filter.json
+++ b/tests/filter.json
@@ -3674,6 +3674,58 @@
       ]
     },
     {
+      "name": "quoted True, double quotes",
+      "selector": "$[?@.a==\"True\"]",
+      "document": [
+        {
+          "a": "True"
+        },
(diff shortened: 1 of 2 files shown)


tests/filter.json
```diff
@@ -3673,6 +3673,52 @@
...
       ]
     },
+    {
+      "name": "quoted True, double quotes",
+      "selector": "$[?@.a==\"True\"]",
+      "document": [
+        {
+          "a": "True"
+        },
+        {
```
f3ath: Would it make sense to add an entry like
```json
{
  "a": true
}
```
and similar ones to the other tests?

f3ath: Thank you!

PR #115, opened 2026-05-17
Fix line endings in all files according to .editorconfig

taimoorzaeem:
The repository settings dictate that all files should have line endings, but this isn't fixed for all files. This is annoying when editing files because git diff shows irrelevant line ending additions. So this fixes it once and for all.


f3ath:
Thank you, @taimoorzaeem

PR #120, opened 2026-09-14
Test slice start is equal to array length

jg-rp:
Test that the slice selector selects no elements when the start index is equal to the length of the input array, with end index and step omitted.

This covers a bug found in https://github.com/jg-rp/json-p3/issues/47. It was accidentally selecting the last array element.


taimoorzaeem (approved):
Works with my implementation. 👍 

Side note: You need to also update `cts.json` file as mentioned in [CONTRIBUTING.md](https://github.com/jsonpath-standard/jsonpath-compliance-test-suite/blob/main/CONTRIBUTING.md#tags).

jg-rp:
Thanks @taimoorzaeem. CI will build `cts.json` automatically.

> [optional] run the build.sh or build.ps1 script located in the root folder
> (this will be performed automatically by GitHub CI after the pull request has been merged to main)

Although I do see that the validation workflow does not build `cts.json` before validating. Nor does the "Build 'cts.json'" workflow validate after building. Maybe that doesn't matter. If all JSON files in `tests/` are valid then we can assume a fresh `cts.json` build is valid.

f3ath:
Thank you!

